PR #7, opened 2016-06-27
Add keyboard navigation commands

clehner:
h/l: scroll through tabs
j/k: scroll through messages
x: close current tab

Depends on dominictarr/hypertabs#1

There may be a way to do this fitting more with the plugs paradigm, but this is a start anyway.

### DIFF
--- a/keyscroll.js
+++ b/keyscroll.js
@@ -1,0 +1,29 @@
+module.exports = function (container) {
+  var curMsgEl
+
+  container.addEventListener('click', onActivateChild, false)
+  container.addEventListener('focus', onActivateChild, true)
+
+  function onActivateChild(ev) {
+    for (var el = ev.target; el; el = el.parentNode) {
+      if (el.parentNode == container) {
+        curMsgEl = el
+        return
+      }
+    }
+  }
+
+  function selectChild(el) {
+    if (!el) return
+    el.scrollIntoViewIfNeeded()
+    el.focus()
+    curMsgEl = el
+  }
+
+  return function scroll(d) {
+    selectChild(!curMsgEl ? container.firstChild
+      : d > 0 ? curMsgEl.nextElementSibling || container.lastChild
+      : d < 0 ? curMsgEl.previousElementSibling || container.firstChild
+      : curMsgEl)
+  }
+}
--- a/modules/feed.js
+++ b/modules/feed.js
@@ -4,6 +4,7 @@ var Scroller = require('pull-scroll')
 var h = require('hyperscript')
 var pull = require('pull-stream')
 var u = require('../util')
+var keyscroll = require('../keyscroll')
 
 var plugs = require('../plugs')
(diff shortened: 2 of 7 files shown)


dominictarr:
okay, merged. okay we should figure out how to make this fit the plugins model.
it ought to be a one line change to enable or disable this plugin.
two ideas:
* tabs to exports `tabs_control` plug, which the keyboard controler fits into.
* there is a "decorator" that wraps the tabs plugin - it would call the current tabs plugin, and then modify or "decorate" it. I havn't actually used any decorator style yet though...

dominictarr:
how can we refactor this so that it's not necessary to add a `.scroll()` property to each column?
because that is the sort of thing which will break if there are more plugins

clehner:
> how can we refactor this so that it's not necessary to add a `.scroll()` property to each column?
because that is the sort of thing which will break if there are more plugins

Refactored in c9b2708a2c9961e39c5c2370d526d929215013df to have the tabs plugin add `.scroll()` to the tabs instead of in each tab. If a tab doesn't have a `.scroller__content` element, the scroll method will do nothing. There may still be a better way...

> tabs to exports tabs_control plug, which the keyboard controler fits into.

That makes sense. I would have to do some more figuring out to see how exactly to do that.

dominictarr:
aha! there is `document.activeElement` which is the currently focused item.
http://stackoverflow.com/questions/497094/how-do-i-find-out-which-dom-element-has-the-focus#497108

can use that to get the active element instead of storing scroll state in a closure. then it isn't necessary to monkeypatch elements created by other plugins.

clehner:
that looks useful - i didn't know about that property. i see it as an advantage to store scroll state stored per-tab. so if i switch between tabs the active message in each tab is maintained. now that i think about it though, maybe the per-tab active element could be recovered approximately by just picking an item that is in view if the active element is not in the active tab. then keyscroll could be more stateless.

dominictarr:
also, this might be useful for detecting when an element is removed: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver

dominictarr:
I know what you mean. we definitely don't want to have columns scroll back to the top (say). it would be acceptable to have some sort of integration that a view provides to enable keyboard control...
like, maybe it would be acceptable to set a class to represent the selected item? the concept of "selecting" is pretty generic.

I've realized I'm setting a bad example re:keyboard control... I have keyboard stuff littered around the place.

but here is one thing which might be a good idea:
https://github.com/dominictarr/patchbay/blob/master/modules/message.js#L44-L49

When you press enter on a message, it queries for an element with ".enter" class, and `.click()`s on it. that way, it's easy for whatever plugin under that (in this case, open message as a new tab) to fire.

Also, I set all the messages to have `tabindex=0` (magic setting so you can tab to them...)

We should probably continue hacking on keyboard control, but look for a pattern to modularize and abstract it later.

PR #8, opened 2016-06-27
WIP: Search & stuff

clehner:
This adds search, including a message view where a path for a search query starts with a '?', and a keybinding of '/' to activate the search field. The search is basically a grep of the log, like how Patchwork and git-ssb-web do it.
Also I did a minor refactor of how keyscroll is used, and other minor changes.

Currently search causes patchwork/sbot's CPU usage to go high for a while, I think because the operation reads the whole log and muxprc doesn't pass through the backpressure. I think I might be able to work around this by using `u.next()` to combine multiple requests with `limit: N` instead of doing an unlimited request and `take(N)`. Also, if a search query returns no results, the use of `u.next()` triggers the query to be repeated after the stream ends, even after the search tab is closed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ now clone and run patchboard.
 git clone https://github.com/dominictarr/patchbay.git
 cd patchbay
 npm install electro electron-prebuilt -g
-patchwork plugins.install ssb-links # must have patchwork >=2.8
+patchwork plugins.install ssb-links ssb-query # must have patchwork >=2.8
 electro index.js
 ```
 
--- a/modules/channel.js
+++ b/modules/channel.js
@@ -1,0 +1,55 @@
+var h = require('hyperscript')
+var u = require('../util')
+var pull = require('pull-stream')
+var Scroller = require('pull-scroll')
+
+var plugs = require('../plugs')
+var message_render = plugs.first(exports.message_render = [])
+var message_compose = plugs.first(exports.message_compose = [])
+var sbot_log = plugs.first(exports.sbot_log = [])
+
+exports.message_meta = function (msg) {
+  var chan = msg.value.content.channel
+  if (chan)
+    return h('a', {href: '##'+chan}, '#'+chan)
+}
+
+exports.screen_view = function (path) {
+  if(path[0] === '#') {
+    var channel = path.substr(1)
+
+    var content = h('div.column.scroller__content')
+    var div = h('div.column.scroller',
+      {style: {'overflow':'auto'}},
+      h('div.scroller__wrapper',
+        message_compose({type: 'post'}), //header
+        content
+      )
+    )
+
+    function matchesChannel(msg) {
+      if (msg.sync) console.error('SYNC', msg)
+      var c = msg && msg.value && msg.value.content
+      return c && c.channel === channel
+    }
+
+    pull(
+      sbot_log({old: false}),
+      pull.filter(matchesChannel),
+      Scroller(div, content, message_render, true, false)
+    )
+
+    pull(
+      /*
+      sbot_query({query: [
+        {$filter: {value: {content: {channel: channel}}}}
+      ]}),
+      */
+      u.next(sbot_log, {reverse: true, limit: 100, live: false}),
+      pull.filter(matchesChannel),
+      Scroller(div, content, message_render, false, false)
+    )
+
+    return div
+  }
+}
--- a/modules/channel.js
+++ b/modules/channel.js
@@ -7,6 +7,7 @@ var plugs = require('../plugs')
 var message_render = plugs.first(exports.message_render = [])
 var message_compose = plugs.first(exports.message_compose = [])
 var sbot_log = plugs.first(exports.sbot_log = [])
+var sbot_query = plugs.first(exports.sbot_query = [])
 
 exports.message_meta = function (msg) {
   var chan = msg.value.content.channel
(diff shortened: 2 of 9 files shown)


dominictarr:
sweet! I have merged this up to 3bf33cc. looking forward to having search.
speaking of search, note that patchbay has autosuggest for images using `!`
it would be cool to have an inline search for mentioning messages like that too... *day dreams*

we'd need proper indexing to make it fast enough though

dominictarr:
oh, could you make keyscroll it's own module and then make a PR so that patchbay depends on it?
it has quite clear functionality - I can easily imagine reusing it elsewhere - so it should be it's own module.

also re: `u.next` I'm thinking of building this idea into muxrpc so everything gets back pressure, but we'll need to wait for multiserver to make the upgrade work.

clehner:
> note that patchbay has autosuggest for images using !
it would be cool to have an inline search for mentioning messages like that too... day dreams

nice. i wonder if suggest-box itself might benefit from using `Scroller`, keyscroll, and streams 

i was also thinking of autocomplete for choosing or navigating to a #channel. what would you think of having channels in patchbay?

i'll pull out keyscroll into a module

dominictarr:
yes yes and yes. Basically the core idea for patchbay is that each tab is mostly that a tab is a just query on the database.

Have been thinking about having a "location bar" that you can type [queries](https://github.com/dominictarr/ssb-query) into, then opening that as a tab...

clehner:
location bar makes sense. being able to put raw queries in it would be interesting.

clehner:
I got around most of the problem with backpressure and searches by filtering the stream after passing it through `u.next` instead of before. I also added channels. There is still a problem that if a channel or search has no messages, the stream gets stuck in a loop at the end where next keeps getting called. I think this is a misbehavior of `u.next`. Also, if a tab is closed and the stream is still trying to load items, it keeps trying to do load items even though it is not going to render thm. It would be better if closing the tab triggered a Scroller sink abort. I suspect exposing the pause method on pull-scroll might help make this work.

I tried using ssb-query to filter the log for channel view, and it was slower than filtering the regular log. However, after I added an index for `[['value', 'content', 'channel'], 'timestamp']`, it became very fast. Getting a list of all channel names (for suggest-box) is still slow though. I'm doing it like this:
```
  pull(
    sbot_query({query: [
      {$map: ['value', 'content', 'channel']},
      {$filter: {$prefix: ''}}
    ]}),
    pull.unique(),
    ...
  )
```
Is there a more efficient way?

dominictarr:
yeah, ssb-query still needs a bit of work, and since it doesn't currently have an index for channel it will be a scan anyway. yup, closing a tab should absolutely abort any streams.

selecting the list of channels should be a reduce.
``` js
  pull(
    sbot_query({query: [
      {$filter: ['value', 'content', 'channel', {$prefix: ''}]}, //filter to posts that have a string for "channel"
      //or replace {$prefix: ''} with {$type: "string"} think that should work?
      {$reduce: {channel: ['value', 'content', 'channel'], posts: {$count: true}}} //posts per channel
    ]}),
   ...
  )
```
if we had an index for channel that would be pretty fast. This is a pretty good test case. I havn't needed to do a query like this yet. it will be possible to materialize/index a reduce too, which would make this lightening fast.

clehner:
The speedup I got for viewing messages in a channel was by adding this to ssb-query's index: `{key: 'cha', value: [['value', 'content', 'channel'], 'timestamp'] },` (I'll make a PR for this later).

This filter gave no results:
```
{$filter: ['value', 'content', 'channel', {$prefix: ''}]}
```
This one gave a TypeError after a while:
```
{$filter: {value: {content: {channel: {$prefix: ''}}}}}
```

Using `{$type: 'string'}`or `[{$type: true}, {$eq: 'string'}]` didn't seem to work.

However, I found that after `{$map: ['value', 'content', 'channel']}`, only strings were left anyway, so it doesn't need the string filter. It's cool that the reduce counts the posts too. Although the disadvantage of using the reduce here is that the results cannot be used incrementally, only once the reduce is done. I understand indexing the reduce would solve that. Can this be done by adding another thing to `indexes` in ssb-query, or would this take more work in streamview-links?

clehner:
I realized that the channel view has messages sorted by timestamp ascending. Is it possible to read a ssb-query in reverse?

clehner:
Never mind, adding `reverse: true` worked :P

dominictarr:
oh, yeah. sorry, arrays have a different meaning in filter queries to map and reduce queries. so that query is wrong.

it will be possible to have a realtime reduce. you could let this read out the currently stored data, and then realtime update after that, or I guess you could set a min latency that you want and just get realtime. we'd have to write `merge` functions to correspond to these: https://github.com/dominictarr/map-filter-reduce/blob/master/simple/reduces.js

and some stuff here to make a aggregate merge function https://github.com/dominictarr/map-filter-reduce/blob/master/make.js#L155-L179

also on the TODO list is a terser query syntax. I'm happy that you can use JSON, but it's pretty verbose and has a lot of parenthesis to match.

dominictarr:
WOW, search responds SO FAST !!!

dominictarr:
merged into 1.2.0, thanks this is awesome!

dominictarr:
hmm, question: so for this, we really want to abbreviate the posts somehow, and highlight the search terms. What are some approaches we could apply to do that within the patchbay module system?

NHQ:
if you implement channel indexes, let there be multiple channels #justSayin #isAll

clehner:
For highlighting search results, we could wrap message_render in search.js with a function that would take the message DOM element and add highlighting. [text-node-searcher](https://github.com/clehner/text-node-searcher) or some other module could do the highlighting. For abbreviating posts I think it would be easier to do it at the DOM level than the HTML level or markdown level. It might be tricky to do that well. So I'm not sure how this would work with the plug system, since it seems particular to the search feature.

@NHQ ssb-query as of v0.1.0 indexes the channel property. i'm not sure how that would handle having channel be an array.

PR #10, opened 2016-06-30
Moar search

clehner:
32f894d: Fixed u.next looping on empty streams <https://github.com/dominictarr/patchbay/pull/8#issuecomment-228944777>
86a4987: Improve speed of channels query
415b47a: Add feed navigation suggestions (press `@` to get a prompt with name/id suggestions)
ba2f9b6: Highlight search terms <https://github.com/dominictarr/patchbay/pull/8#issuecomment-229595810>

### DIFF
--- a/modules/search-box.js
+++ b/modules/search-box.js
@@ -40,8 +40,11 @@ exports.search_box = function (go) {
 
   pull(
     sbot_query({query: [
-      {$map: {channel: ['value', 'content', 'channel']}},
-      {$reduce: {channel: 'channel', posts: {$count: true}}}
+      {$filter: {value: {content: {channel: {$gt: ''}}}}},
+      {$reduce: {
+        channel: ['value', 'content', 'channel'],
+        posts: {$count: true}
+      }}
     ]}),
     pull.collect(function (err, chans) {
       if (err) return console.error(err)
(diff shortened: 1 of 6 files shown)


dominictarr:
merged into 1.4.0
I added a bit to sort the suggested names. we should put the suggest query stuff in it's own module, it's used by compose too, it seems reasonable that someone might want a custom suggestion algorithm

clehner:
I saw the suggests plug used in the composer but I didn't use it because I see the suggestions for the search box as different from the suggestions for the composer. In the search box, completing a name/feed should give you `@feed`, while in the composer it expands to `[@name](@feed)`. I also explored a different approach for loading the suggestions for the search box. The channels and feeds/users are loaded at app start and given to `suggest-box` as an array. When `suggest-box` gets an array it filters and sorts them internally, so pre-sorting the array does not have an effect. This is different from the approach used in the composer suggestions which is to run a query filtering with `$prefix` and give the sorted results to `suggest-box` asynchronously.

I wanted to have the channels and feed name suggestions in the search box appear quickly, which is why i let it pre-fetch the data (like in Patchwork). The downsides are that it takes more CPU to run the queries when the app starts, and the info is not updated as new messages come in. Indexing reduces with live updates as you discussed would make this better.

The search box suggestions could probably be changed to use the suggest plug and asynchronous query which would be slower but probably tolerable, but I'm not sure what to do about the completions for feeds expanding to markdown in the search box.

dominictarr:
yeah. this is great for now. we can update this once we have realtime reduces.

Same with the suggest plugin. We can probably separate that into several parts, the ui (which composer and search will have their own) and then the queries, which return object streams.

dominictarr:
I bet if it was just one async round trip (without redoing the whole query) then async lookup would be acceptable.

PR #12, opened 2016-06-30
search respects word boundries and multiple terms

dominictarr:
@clehner more search stuff.

matches on exact word boundries and supports multiple search terms.
this makes search quite a bit more useful.

i'm not 100% sure about using exact word boundries, but it should probably be the default.
I tried to search for number5's post about address-something. I rememebered the word "address",
but i didn't remember it was "addressable" maybe the answer is globbing? `address*`?

or maybe stemming? but is there a language agnostic way to do stemming? (guessing maybe with markov chains?)

I'm quite impressed with how well brute force search is working!

Also, I'm noticing that feeds + keyboard scroll, and suggest + keyboard select is very similar. we could make these all use the same modules, and if we had a compact way to render a message, have suggest-for-messages in composer too.

### DIFF
--- a/modules/search.js
+++ b/modules/search.js
@@ -27,15 +27,21 @@ function searchFilter(terms) {
     return c && (
       msg.key == terms[0] ||
       andSearch(terms.map(function (term) {
-        return new RegExp(term, 'i')
+        return new RegExp('\\b'+term+'\\b', 'i')
       }), [c.text, c.name, c.title])
     )
   }
 }
 
+function createOrRegExp(ary) {
+  return new RegExp(ary.map(function (e) {
+    return '\\b'+e+'\\b'
+  }).join('|'), 'i')
+}
+
 function highlight(el, query) {
   var searcher = new TextNodeSearcher({container: el})
-  searcher.setQuery(query)
+  searcher.query = query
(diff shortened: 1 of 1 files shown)


clehner:
i'm okay with search terms using word boundaries, since i can still use `.*word.*` to search within words. what do you mean by stemming? also, you know you can construct a regex to match multiple terms with OR by joining the terms with `|`? if you put a regex on a TextNodeSearcher instance using `searcher.query = regex` instead of `searcher.setQuery(str)`, only one searcher would be needed, so it wouldn't have to touch the DOM as much. of course these are minor optimizations that probably won't matter.

dominictarr:
good points. fixed.
stemming is a natural language processing technique where you take different versions of a word and transform back to a base version. so that you can search for "address" and also detect things matching: addresses, adresser, addressee, addressable, addressability

dominictarr:
I tested `.*` and matched an entire paragraph. regular expressions are powerful, but they are too powerful most of the time. I'd lean towards having a simplified globbing pattern with `*word*` and then if you want full blown regular expressions, use `/.../`

dominictarr:
merged into 1.5.0

PR #13, opened 2016-07-01
CSS tweaks

evbogue:
This is a few minor tweaks of the CSS from earlier. Also I worked on the profile pages a little bit.

Two things to note:

The render on the search bar suggestions still runs into the right side. I did a margin hack to knock it over a bit, but that isn't ideal. The search bar might need to go back on the left side, even though that doesn't seem as balanced.

Also the module hyperlightbox is overwriting styles on the lightbox, so I can't write to them locally. Ideas?

### DIFF
--- a/modules/avatar-profile.js
+++ b/modules/avatar-profile.js
@@ -6,7 +6,7 @@ var avatar_action = plugs.map(exports.avatar_action = [])
 
 exports.avatar_profile = function (id) {
 
-  return h('div.row',
+  return h('div.row.profile',
     avatar_image(id),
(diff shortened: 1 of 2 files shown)


dominictarr:
what do you want to change about lightbox?
maybe there is another way for lightbox to do what it needs?

dominictarr:
okay update to hyperlightbox@0.1.0 - now it only sets fixed boundries, but doesn't set background color or border.

dominictarr:
oh oops, i didn't actually intend to merge this, yet. but whatever. it's okay.
lightbox is a bit inbetween now.

PR #15, opened 2016-07-01
Right align message metadata

dominictarr:
I think this looks better.
screenshot_right-align-meta

i think the next thing is the ID input is ugly. That could be a "copy id" option like patchwork has,
or also there could be a header that appears on the message view (i.e. the thread page) which would have a selectable input box.

### DIFF
--- a/modules/avatar-profile.js
+++ b/modules/avatar-profile.js
@@ -6,7 +6,7 @@ var avatar_action = plugs.map(exports.avatar_action = [])
 
 exports.avatar_profile = function (id) {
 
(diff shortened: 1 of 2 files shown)


dominictarr:
@evbogue @clehner what do you think?

evbogue:
@dominictarr Looks good to me! 

I agree that the id could be hidden or more beautiful than it is.

clehner:
i like

dominictarr:
merging.

dominictarr:
merged into 1.5.1

PR #16, opened 2016-07-02
Error handling

clehner:
- Fixed some errors.
- Added a `errors` tab to show new errors. Is this too developer-centric?

### DIFF
--- a/modules/compose.js
+++ b/modules/compose.js
@@ -57,7 +57,7 @@ exports.message_compose = function (meta, prepublish, cb) {
     }
     message_confirm(meta, function (err, msg) {
       ta.value = ''
-      cb(err, msg)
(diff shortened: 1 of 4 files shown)


dominictarr:
hmm. as a developer, I think I'd prefer to get errors in the dev tools?
does this interfeer with that?

clehner:
It shouldn't interfere with it, no

dominictarr:
okay so ev.preventDefault() means that errors arn't in the console anymore. can we remove that?

clehner:
yes, i think it's safe to remove that. sorry, i tried throwing an error in a timeout and it showed up in the console so i assumed it was fine.

dominictarr:
I removed this at some point, but this still completely swallows errors in firefox

PR #19, opened 2016-07-02
Search header

dominictarr:
add a header to the search page to show how many messages have been examined.

screenshot-search-fish


dominictarr:
updates in realtime.

clehner:
nice. should the header stay fixed when scrolling through the results? currently it goes out of view when you start to scroll down.

dominictarr:
@clehner yeah I was thinking about that. maybe it could be on a status panel attached to say, the bottom left corner?

PR #21, opened 2016-07-03
Firefox

dominictarr:
These changes make it basically work in firefox.
some more work is needed, like allowing peers to connect to a remote pub.
via
https://github.com/dominictarr/ssb-ws
and
https://github.com/dominictarr/ssb-lite

Also, some of the tabs (in particular, /private) doesn't really work great.
but we can definitely have a MVP light client working very soon.


dominictarr:
merging this.

dominictarr:
merged into 1.5.3

PR #25, opened 2016-07-06
show current location in search-box

dominictarr:
This change shows the current location in the search box when switching tabs.

This means you can always copy the id of a message or feed from here,
which means message-id can be removed, which looks much cleaner.

I just added a feature to hypertabs to take a listener called when a new tab is selected.


dominictarr:
merged into 1.6.0

PR #26, opened 2016-07-07
Remember selected tab

clehner:
Use sessionStorage so that the selected tab is remembered if you refresh the page, but if you restart the electron app you still get the default public view


dominictarr:
awesome, had been wanting this!
merged into 1.6.0

PR #20, opened 2016-07-07
Notifications

clehner:
Would we like to see some kind of notifications in Patchbay of mentions and things related to the current uxer?

Possible implementation approach: factor out notifications from ssb-notifier and add a patchbay tab to show them like a feed. This would not take into account watched threads like in Patchwork.

### DIFF
--- a/modules/index.js
+++ b/modules/index.js
@@ -17,6 +17,7 @@ module.exports = [
   require('./message-name.js'),
   require('./message.js'),
   require('./names.js'),
+  require('./notifications.js'),
   require('./post.js'),
   require('./private.js'),
   require('./public.js'),
--- a/modules/notifications.js
+++ b/modules/notifications.js
@@ -1,0 +1,114 @@
+var h = require('hyperscript')
+var u = require('../util')
+var pull = require('pull-stream')
+var Scroller = require('pull-scroll')
+var paramap = require('pull-paramap')
+var plugs = require('../plugs')
+
+var message_render = plugs.first(exports.message_render = [])
+var sbot_log = plugs.first(exports.sbot_log = [])
+var sbot_whoami = plugs.first(exports.sbot_whoami = [])
+var sbot_get = plugs.first(exports.sbot_get = [])
+var message_unbox = plugs.first(exports.message_unbox = [])
+
+function unbox() {
+  return pull(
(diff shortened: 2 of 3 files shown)


clehner:
Attached a commit adding a notifications tab. It is very basic, it just filters messages from the log based on whether they have something to do with the current user. It doesn't use code shared with ssb-notifier but it uses similar logic to detect mentions, follows, digs, and replies. For now it just renders the messages as-is, not giving more detailed descriptions of their relevance to the user.

dominictarr:
merged into 1.6.0

dominictarr:
This is a good start. I'd like to be able to separate things I will react to like replies and mentions from things that are merely informational, like likes.

clehner:
>  I'd like to be able to separate things I will react to like replies and mentions from things that are merely informational, like likes.

That makes sense. I was thinking similarly about likes taking up a lot of space in the public feed.

Some ideas:
- hide likes from notifications
- put likes of your stuff into a separate tab
- decrease the size and/or visual prominence of likes (and follows). i.e. make them fit on one line, without border/bg, with smaller avatar icon
- combine multiple like messages into a single item in the feed

dominictarr:
I was thinking an easy thing to do, while maintaining the streamyness (which is a key part of the architecture) would be to reduce "simple actions" (likes, follows, etc) into a single item.
it would just say "bob yupped X,Y,Z and followed A,B,C" in one line. Ususally you get a bunch of reactions from one person, because they come on, interact with a bunch of stuff, then stop.

dominictarr:
oh, question: does notifications do a full scan? for new users on a lite client we don't want to make that download the whole feed. It should probably stop once it gets to messages earlier than their first message.

clehner:
> does notifications do a full scan?

yes.

> for new users on a lite client we don't want to make that download the whole feed. It should probably stop once it gets to messages earlier than their first message.

agreed. i'll add that

> I was thinking an easy thing to do, while maintaining the streamyness (which is a key part of the architecture) would be to reduce "simple actions" (likes, follows, etc) into a single item.
it would just say "bob yupped X,Y,Z and followed A,B,C" in one line. Ususally you get a bunch of reactions from one person, because they come on, interact with a bunch of stuff, then stop.

that would make sense for `public` and `feed`.

PR #28, opened 2016-07-09
WIP: Invites

dominictarr:
default should be over websockets (using multiserver) so that they work from the browser.
this also upgrades to ssb-client@4 which is rewritten in terms of multiserver.

it's pretty ugly right now. need to confirm that it actually does work from the browser.
should do though.

### DIFF
--- a/modules/index.js
+++ b/modules/index.js
@@ -26,6 +26,7 @@ module.exports = [
   require('./suggest.js'),
   require('./tabs.js'),
   require('./thread.js'),
-  require('./timestamp.js') 
+  require('./timestamp.js'),
+  require('./invite.js')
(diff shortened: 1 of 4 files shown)


dominictarr:
merged into 1.7.0

clehner:
does this take currently-existing invites, or is there a new API for making invites to use here?

dominictarr:
no, sorry. It uses a multiserver syntax for a code.

it should be:
`ws://<host>:<port>~shs:<pub_key>:<seed>`

using `ssb-ws` port will be 8989. you need to install ssb-ws on your pub to receive these.
and currently i've been assembling these manually.

PR #30, opened 2016-07-10
Keep timestamps up-to-date

clehner:
Once a minute re-render all timestamps on the page.
Show full datetime string in `title` tooltip.
This makes it easier to tell when things were posted.


dominictarr:
merged into 1.7.0

PR #33, opened 2016-07-11
Optimize notifications stream

clehner:
Abort reading the log past the oldest message in your feed, since presumably no one will have mentioned your feed before you started posting on it.

Also fix an error having to do with messages where the branch property is an array


dominictarr:
merged into 1.7.1

PR #35, opened 2016-07-12
Fixes

clehner:
- Fix message links and word word space in about messages
- Update ssb-avatar dependency


dominictarr:
thanks! merged into 1.7.4

PR #41, opened 2016-07-19
yups to digs, digs to bottom right corner

evbogue:
I changed yups to digs for consistency across the three apps, Patchwork/git-ssb-web/Patchbay. Feel free to reject if yup is preferred. 

In the future perhaps the deployer can set this to 'Star' or 'Yarr' in a config file?

WIP commit. Next I'm going to try to get voting into the bottom right corner of the screen. Next next I'm going to try to get a reply button down there too.

Ok this is mostly accomplished. The one goal I have is to remove the ' | ' when there is no dig button.

### DIFF
--- a/modules/like.js
+++ b/modules/like.js
@@ -27,7 +27,7 @@ exports.message_meta = function (msg, sbot) {
     pull.collect(function (err, votes) {
       if(votes.length === 1)
         digs.textContent = ' 1 Dig'
-      if(votes.length)
+      if(votes.length > 1)
         digs.textContent = ' ' + votes.length + ' Digs'
     })
   )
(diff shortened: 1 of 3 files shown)


evbogue:
Ok this is mostly accomplished. The one goal I have is to remove the ' | ' when there is no dig button.

dominictarr:
I kinda like yup. but that is why I put the "expression" field on.
http://wx.larpa.net:8989/msg/%ym8xH+z4R9qId5ioPV5jALGM0m2Kp4EPwt6bEM+kQGY=.sha256

I think the _user_ should be able to choose their own expression... because language needs to evolve from the bottom up, from use...

What does the `|` do? does it just mark the end of the actions?

evbogue:
The | (pipe) was simply a symbol between Dig/Yup and Reply. It seemed to balance it out a bit visually.

"Dig | Reply"

Bottom-up expressions seems to be a good idea. Also it'd allow for drop-downs and such.

evbogue:
I don't particularly love 'Dig' either, as it reminds of the Digg days and also seems to be 1960's hippy slang. It only seemed to me that people might have been confused about what a Yup is.

PR #45, opened 2016-07-20
Split view

dominictarr:
this adds basic support for split views.

screenshot_patchbay_split

as you'll notice from looking at the commits, this was very simple to implement!
since we had previously made sure that each module was well behaved, and since we used flex box, and the width is reactive... it just worked! ;)

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,8 +21,22 @@ var u = require('./util')
 //var i = indexes.indexOf('index.js')
 //indexes.splice(i, 1)
 
-modules['app.js' ] = {app: app}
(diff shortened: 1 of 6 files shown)


evbogue:
The return of the 'deck! 

This will be awesome with a /contacts view and /public for if/when there are too many people in public to parse the data.

dominictarr:
yeah, we don't currently have a way to view say, followers of a feed, which this would be useful for.

dominictarr:
merged into 1.8.0

PR #47, opened 2016-07-21
make private messages message approachable, uppercase publish

gwenbell:


dominictarr:
thanks, merged into 1.8.2

PR #48, opened 2016-07-21
add space and separator between Dig and Reply

evbogue:
This is what I meant to accomplish yesterday with a '|' character. I figured out that it's easier to accomplish with css!

2016-07-20-232754_607x512_scrot


dominictarr:
thanks! merged into 1.8.1

PR #49, opened 2016-07-21
separate markdown into it's own plugin, and make blobs work on liteclient

dominictarr:
implemented a bunch of things to make liteclients work better.
if you don't have ssb-links installed on your pub, then avatars will break for liteclients that currently work


dominictarr:
merged into 1.8.0

PR #51, opened 2016-07-21
stabbing in the dark

cryptix:
This adds two approaches for adding music to patchbay:

* `music-release-cc.js` displays `type: music-release-cc` posts (like `%e/JG0Stf90f4FBpBSPC+OAwl5EAIL5Xz3LFz0mhgpl8=.sha256`)
* `music-relase.js` and friends which depend on (`root`,`branch`)  linked `meta-image`, `audio-mp3` type posts to aggregate a release. I linked some examples of this in `%KgIliFBy5MeXafVENTNkRVHdFTy+d+T2zGSy9EuSYBg=.sha256`)

Visually I like the first approach more but dislike the _fat_ objects it uses. The 2nd approach gives you the option to dig/pin/star/whatever individual tracks/meta and recompile them in playlists for instance.

### DIFF
--- a/modules/music-release-cc.js
+++ b/modules/music-release-cc.js
@@ -1,0 +1,78 @@
+var markdown = require('ssb-markdown');
+var h = require('hyperscript');
+var u = require('../util');
+var ref = require('ssb-ref');
+
+//render a message
+
+var plugs = require('../plugs');
+var message_link = plugs.first(exports.message_link = []);
+var message_confirm = plugs.first(exports.message_confirm = []);
+var sbot_links = plugs.first(exports.sbot_links = []);
+
+exports.message_content = function(msg, sbot) {
+    if (msg.value.content.type !== 'music-release-cc')
+        return;
+
+    var tracks = msg.value.content.tracks;
+    return h('div',
+        h('img', { "src" : "http://localhost:7777/" + encodeURIComponent(msg.value.content.cover) }),
+        h('h1', msg.value.content.title),
+        h('ol',
+                 Object.keys(tracks).map(function(k) {
+                     var t = tracks[k];
+                     return h('li', t.fname,
+                         h("br"),
+                         h('audio', {
+                                  "controls" : true,
+                                  "src" : "http://localhost:7777/" + encodeURIComponent(t.link)
+                              }))
+                 })),
+        h('p',
+                 "More info:", h('a', { href : msg.value.content.archivedotorg }, "archive.org"),
+                 h("br"),
+                 "License:", h('a', { href : msg.value.content.license }, "Link")))
+}
+
+// copied from like.js
+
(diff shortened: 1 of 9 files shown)


cryptix:
rebased

modules/music-release-cc.js
```diff
@@ -0,0 +1,78 @@
...
+                     var t = tracks[k];
+                     return h('li', t.fname,
+                         h("br"),
+                         h('audio', {
+                                  "controls" : true,
+                                  "src" : "http://localhost:7777/" + encodeURIComponent(t.link)
+                              }))
+                 })),
+        h('p',
+                 "More info:", h('a', { href : msg.value.content.archivedotorg }, "archive.org"),
```
dominictarr: Instead of a hard coded link to archive.org, you could just have a list of references (or a markdown section). That would be more runcible.

music-release.json
```diff
@@ -0,0 +1,8 @@
+{
+    "type": "music-release",
+    "title": "(miniatura070) bleupulp - the fall of",
+    "archivedotorg": "https://archive.org/details/Miniatura070Bleupulp-theFallOf",
+    "license":"https://creativecommons.org/licenses/by-nc-nd/3.0/",
+    "cover": "&2JGx5HkFts0PS84K3xwKGZZgqQ5zAf2LSuLAahBhm6I=.sha256",
+    "ipfs": "/ipfs/QmSuUXxGsF7yQNPDQEjUsAmWZZoNNmdMmZK59MsQroX3ch"
+}
```
dominictarr: hmm, these json files did not need to be checked in.

dominictarr:
Okay this is merged into 1.13.0
This does need to be tidied up a bit. my main concern is that it creates quite a few files.
I think you should choose one approach and go with it. personally I prefer the fat object approach,
because it's more self contained. you could still include something in other compilations by linking to the blob, and referencing the source (or use backlinks to show it also appears in these other collections)

I'm guessing you made the files manually (since you left some json files there) I guess next step is a tool to create a set for you. A command line tool is still progress.

Oh, and a playlist queue thing ;)

my policy for patchbay is gonna be merge everything I can (I don't want to have repeat of editable-posts debacle) until I can't stand it anymore and then implement the ideas I have for on-the-fly plugins (it should be fairly straightforward to port any patchbay plugin to that)

music-release.json
cryptix: Oh.. yea, sorry about that.
I could have documented these inside the rendering js files as comments.

modules/music-release-cc.js
cryptix: This was really just prototyping and a way to attribute the source. 

How would you edit these, though? It's also why I shunned away from the fat objects...
We didn't even touch song metadata yet.

PR #56, opened 2016-08-14
misc fixes

mixmix:
patchbay wouldn't start for me. 

there were a couple of things missing, which this PR tries to resolve : 
- a missing module
- some weird css requiring

the latter I fixed in a pretty hacky way ... not sure how it was working before...


dominictarr:
sorry, adding the module is correct, but the `style.css.json`  should be generated by `npm run build` command.

dominictarr:
it would be better not to need the build step when running inside electron... but tricky bit is sometimes we use a browserify build in electron, so you'd have to test whether you where not in the build, and then load the css via the file system...

clehner:
what was wrong with using brfs?

dominictarr:
hmm, that made https://github.com/dominictarr/browselectrify more awkward, i think.

dominictarr:
it was hard to use brfs, _and_ allow fallbacks to actually use the fs module when you are in electron.

mixmix:
@dominictarr feel free to do whatever with this PR - close it if it's served its purpose

PR #64, opened 2016-12-23
A step toward modular CSS (and so more modular components).

ahdinosaur:
%CcTHlP33vXa9mA/lf8dxkUrZaqXA1DcXyZeRVHNb7Z4=.sha256

[`git.scuttlebot.io` link](http://git.scuttlebot.io/%25CcTHlP33vXa9mA%2Flf8dxkUrZaqXA1DcXyZeRVHNb7Z4%3D.sha256)


dominictarr:
is this ready to go? can you describe the changes?

evbogue:
I tested this branch in the lite client, and I'm getting `setImmediate is not defined`. I'm also concerned that the fs.readFileSync will also not work in the lite client.

dominictarr:
If we built it `brfs` we can make `fs.readFileSync` however I would prefer to make mcss compile to javascript so it can be loaded with require, because that will some plans I have down the road a bit easier.

ahdinosaur:
yes, this is more or less ready.

> each module that wants to export styles can either export `css` (a function that returns a css string) or `mcss` (a function that returns a `micro-css` string). these are then merged together in a `styles` module, which is then included to the document head in the `app` module.
>
> for themers, it's only necessary to override the `styles` module and change how the styles are merged together, either with a provided module-override-map or a full complete stylesheet (to match class names in modules).
>
> also chose to keep the current stylesheet the same, so this can be a small change to get the ball rolling in this direction.

also:

- moved backlinks code into a separate module
- changed timestamp export to not be `message_meta`
- updated about messages (added rendering of `description`), also add mini renderer
- add `message_author` which renders avatar, name, and timestamp, also add size option when you want to minify (not sure about `size`, just trying to avoid a `mini` boolean option when there are many possible sizes)

and then probably the most worth discussion: changed the layout of message. originally @mmckegg, @mixmix, and i split this into header, content, and footer, where header contained the author (avatar, name, timestamp) and meta in a flex row, and footer contained actions and backlinks. but then later that night i changed it so everything was in the same container, with author float left, meta float right, then content, actions, and backlinks.

it looks like this:

patchbay message layout

conceptually, it's this:

```txt
------------------------------------------------------
|  author  |           content            |  meta     |
|----------                               ------------|
|                                                     |
|                                                     |
|-----------------------------------------------------|
|                      actions                        |
|-----------------------------------------------------|
|                     backlinks                       |
|-----------------------------------------------------|
```

however now i wonder if i should make the top a separate module that you can add to (like adding "re: blah blah blah" out of the content into that section). same goes with maybe splitting the bottom into three sections, or something.

thoughts?

mixmix:
initial thought before I look at code: that box model is going to break bad for long names - you can already see it making the content width variable for the different names in the example. Might be avoidable with styling but not sure that should be the solution?

modules_basic/index.js
```diff
@@ -12,6 +12,8 @@ module.exports = {
   "invite.js":  require('./invite.js'),
   "like.js":  require('./like.js'),
   "markdown.js":  require('./markdown.js'),
+  "message-author.js":  require('./message-author.js'),
+  "message-backlinks.js":  require('./message-backlinks.js'),
   "message-link.js":  require('./message-link.js'),
   "message-name.js":  require('./message-name.js'),
   "message.js":  require('./message.js'),
```
mixmix: would prefer message to be above the message partials ... 
notice in this is starting to feel like it could like a folder....

modules_basic/message-author.js
```diff
@@ -0,0 +1,39 @@
...
+    message_author,
+    mcss: () => fs.readFileSync(Path.join(__dirname, 'message-author.mcss'))
+  }
+
+  function message_author (msg, opts = {}) {
+    var { size = 1 } = opts
+    var { value } = msg
+    var { author } = value
+    return h('MessageAuthor', {
+      className: size === 0 ? '-mini' : ''
```
mixmix: is this your opinion? 
are there a range of sizes here ? If not would prefer a boolean opts, like `isMini`

modules_basic/message-author.mcss
```diff
@@ -0,0 +1,12 @@
+MessageAuthor {
+  display: flex
+  flex-direction: column
+
+  -mini {
+    flex-direction: row
```
mixmix: I'm really like this modification pattern

modules_basic/message.js
```diff
@@ -1,122 +1,86 @@
...
-  return function (msg, sbot) {
-    var el = api.message_content_mini(msg)
-    if(el) return mini(msg, el)
+  function message_render (msg, sbot) {
+    var content = api.message_content_mini(msg)
+    if (content) return mini(msg, content)
 
-    var el = api.message_content(msg)
-    if(!el) return mini(msg, message_content_mini_fallback(msg))
+    var content = api.message_content(msg)
```
mixmix: second declaration of variable (remove var)

modules_basic/message.js
```diff
@@ -1,122 +1,86 @@
...
-          msg.querySelector('.enter').click()
-        }
-      }}
-    )
-
-    // ); hyperscript does not seem to set attributes correctly.
-    msg.setAttribute('tabindex', '0')
-
-    return msg
+  function mini(msg, el) {
```
mixmix: this pattern feels a bit weird. I had to read it a couple of times, this feels like a template.
I wonder if it would be good to have a `miniTemplate` and a `standardTemplate` both extracted so that we could see this pattern more clearly?

modules_basic/reply.js
```diff
@@ -0,0 +1,9 @@
+var h = require('../h')
+
+exports.gives = 'message_action'
+
+exports.create = function () {
```
mixmix: other creates take in api even though they don't use them. I think we should be consistent to help other people see the pattern

mixmix: I wonder about an empty needs as well, if we're being super explicit

modules_basic/timestamp.js
```diff
@@ -1,28 +1,47 @@
...
+}
 
 exports.create = function () {
-
-  function updateTimestampEl(el) {
-    el.firstChild.nodeValue = human(new Date(el.timestamp))
-    return el
-  }
-
   setInterval(function () {
```
mixmix: right, this is probably where mutation is being introduced - and needs setImmediate

modules_basic/timestamp.js
```diff
@@ -1,28 +1,47 @@
...
-    var els = [].slice.call(document.querySelectorAll('.timestamp'))
+    var els = [].slice.call(document.querySelectorAll('.Timestamp'))
```
mixmix: oh this makes me feel weird / sad

mixmix: it feels like it breaks some rule... like effects outside of the depject flow.
something to watch

modules_core/app.js
```diff
@@ -1,12 +1,18 @@
...
-  needs: {screen_view: 'first'},
+  needs: {
+    screen_view: 'first',
+    styles: 'first'
+  },
   gives: 'app',
   create: function (api) {
     return function () {
-      document.head.appendChild(h('style', require('../style.css.json')))
+      process.nextTick(function () {
```
mixmix: I'm a noob to this, but there might be a better way to do this if we have setImmediate around ? 
Might be worth checking how Matt did this if you haven't already

modules_core/styles.js
```diff
@@ -0,0 +1,34 @@
...
+var Path = require('path')
+
+// TODO distribute these styles across all
+// the relevant modules, not as a core style.
+var coreStyle = fs.readFileSync(Path.join(__dirname, '../style.css'))
+
+module.exports = {
+  needs: {
+    mcss: 'map',
+    css: 'map'
```
mixmix: guess  I wonder whether we should add css as an option - I think if we're doing mcss, lets stick to that and be pushed around by the opinions it has. I wouldn't want to see people adding piecemeal css just _because it's quicker_

package.json
```diff
@@ -20,9 +22,11 @@
     "hyperprogress": "0.1.0",
     "hyperscript": "^1.4.7",
     "hypertabs": "^3.0.0",
+    "insert-css": "^2.0.0",
     "is-visible": "^2.0.4",
     "kvgraph": "^0.1.0",
     "map-filter-reduce": "^3.0.1",
+    "micro-css": "^0.6.2",
```
ahdinosaur: i've only been using chrome or electron, which has `setImmediate`. i think it's a dependency of `mutant`.

modules_basic/index.js
ahdinosaur: yeah  it _is_ an alphabetical folder listing. :wink: so could replace with `bulk-require` but another pull request.

modules_basic/message-author.js
ahdinosaur: yeah, "add size option when you want to minify (not sure about size, just trying to avoid a mini boolean option when there are many possible sizes)". boolean opts like `isMini` are the worst, when really we want either enumerated sizes or a range of sizes (i like the range between 0 and 1). probably enumerated is better, will change `size` to a string.

modules_basic/reply.js
ahdinosaur: oh, i thought i saw other modules that didn't need anything not receiving api. i'm happy either way.

modules_core/app.js
ahdinosaur: what do you mean? Matt also used `insert-css`, the `process.nextTick` is because of how depject is currently written. see https://github.com/dominictarr/depject/pull/2 for a pull request that would remove the need for `process.nextTick`.

modules_core/styles.js
ahdinosaur: you sure? it seems wrong to me to not give the option of plain css.

modules_basic/timestamp.js
mmckegg: Yeah, I really dislike this. But mostly just because it is not very mutanty - looks a lot more like jquery.

There are much better ways to do this. For example with a hook. Could still be depjected.

ahdinosaur: agree, keen to learn how we could do this better.

package.json
mixmix: yeah it is. can we add the setimmediate package ?

modules_basic/message-author.js
mixmix: not sure what 'change size to string' means ... you mean `size: mini` ? I think that might make sense

modules_core/app.js
mixmix: ok, I've not looked at the details. trust your knowledge on this

modules_core/styles.js
mixmix: I dunno, I was trying a strong stance. I reckon we could try a strong stance (maybe with the css lines commented out to start with)
Feels like a case where it might be good and bringing it back from css to mcss might be a pain

ahdinosaur:
made changes, @evbogue can you see if this works for you now?

evbogue:
The lite client builds now.

mixmix:
hey @ahdinosaur did you sync this with ssb recently ? 

I tried making some changes to padding in response to ahhh arj's padding feedback and got some good looking stuff going on BUT ... the flex with longer content looks like it's not going to work : 

here's a tame but kinda broke example: 
selection_146

here's a well broken one (in terms of readability) : 
selection_147

ahdinosaur:
@mixmix that seems to be expected behavior given the box model, no?

that's the inspiration for my comment:

> however now i wonder if i should make the top a separate module that you can add to (like adding "re: blah blah blah" out of the content into that section). same goes with maybe splitting the bottom into three sections, or something.

so the box model becomes

```txt
------------------------------------------------------
|  author  |          frontlinks          |  meta     |
|---------------------------------------- ------------|
|                      content                        |
|-----------------------------------------------------|
|                      actions                        |
|-----------------------------------------------------|
|                     backlinks                       |
|-----------------------------------------------------|

````

would this address the issue you have in mind?

**Comment by mixmix** [comment]
the styling in those screenshots was in this branch  `about_refactor_styling` on ssb

**Comment by mixmix** [comment]
@ahdinosaur yeah I reckon it could. 

this feels like that structural opinionated edge. could also be : 

```
|-----------------------------------------------------|
|  author  |          frontlinks          |  meta     |
|          |----------------------------- ------------|
|          |           content                        |
|          |------------------------------------------|
|          |           actions                        |
|          |------------------------------------------|
|          |          backlinks                       |
|----------|------------------------------------------|
```

**Comment by mixmix** [comment]
I was wondering if the box model could be 

```
- Message (flex flex-direction: row)
  - Author
  - Other (flex flex-direction: col)
    - Info (flex flex-direction: row)
      - FrontLinks
      - Meta
    - Content
    - Actions
    - Backlinks
```

That way we people could flick the flex direction to column and it would fall back to what you had ?  (This is a naive guess, might not work like that with flex ...)

**Comment by mixmix** [comment]
boom this : 

```
 ------------------------------------------------------
| ---------- ------------------------------------------ |
||  author  | ----------------------------- ---------- ||
||          ||         frontlinks          |  meta    |||
||          | ----------------------------- ---------- ||
||          |------------------------------------------||
||          |           content                        ||
||          |------------------------------------------||
||          |           actions                        ||
||          |------------------------------------------||
||          |          backlinks                       ||
| ---------- ------------------------------------------ |
 ------------------------------------------------------- 
```

**Comment by ahdinosaur** [comment]
yeah, i was trying to avoid nesting as it intuitively feels more rigid, but maybe that's not true. i also wasn't using flexbox, because i couldn't figure out how to make it work without adding container elements (as we had originally at Art~Hack).

anyways, i gave a go at [implementing my idea](https://github.com/ssbc/patchbay/commit/b3f540ed553d092084367d6ea9ed825fcb020048) (except `frontlinks` became `title`): [`dinostyle-refactor`](https://github.com/ssbc/patchbay/compare/dinostyle-refactor)

snapshot

**Comment by ahdinosaur** [comment]
what i mean by "not nested" is [this](https://github.com/ssbc/patchbay/blob/b3f540ed553d092084367d6ea9ed825fcb020048/modules_basic/message.js#L38-L50):

```txt
- Message
  - Author (float: left)
  - title (float: left)
  - meta (float: right)
  - content (clear: both)
  - action
  - Backlinks
```

it felt like minimum viable opinion.

dominictarr:
what if the avatar image floats, but the name is on the top? I think that might be the best of both?

mixmix:
@ahdinosaur it feels like not that MVP, in that the nice flat strucutre doesn't feel like it will allow for much useful styling (to me). Maybe it's possible with CSS wizardry, but I feel less excited by that.

Maybe you could tell us a little more about your motivation for such a flat box-model structure ?

mixmix:
@ahdinosaur here's a possible solution with styling ... if negative margin is ok.

selection_148

It breaks the mini messages but I think that's probably fixable.
pushing this to github on branch `mixstyle-refactor`

ahdinosaur:
@dominictarr can you explain more why you want the name on top?

as opposed to:

```txt
[image] | title, message meta
 name   | ------------------
 time   | content
 ---------------------------
 backlinks, actions
```

@mixmix i like that, can you push up your branch so i can see what you did?

evbogue:
It could be that the position of the date _is_ opinionated. Which may mean we need to leave that choice up to the developer of an individual Bay-based sbot client.

mixmix:
@ahdinosaur pushed ! 

got something working with mini messages. 
I notice the mini message might feel better if it was MessageMini rather than Message -mini

it feels like I was over-riding too much maybe... something to watch 

selection_149

ahdinosaur:
thanks @mixmix, that looks great. :smiley_cat: i feel -mini is working exactly as intended. _shrug_

@evbogue yep, to override this you'd want to override `MessageAuthor` to not include it and use the timestamp module somewhere else, like export as a meta.

arj03:
Really digging the mixmix refactor branch. Only wierd thing I noticed besides long names, is the private messages that looks a bit wierd.

mixmix:
new pull request opened to merge changes into this branch

arj03:
Seems like the latest version doesn't run well with lite client

mixmix:
thanks for the heads up @arj03 . 
How are you running the lite client (I've obviously not looked at that side of the project), and what browser are you using? I trust you `npm install`ed after pulling, there are a couple of new modules?

arj03:
I'm getting these two errors:

insert-css: You need to provide a CSS string. Usage: insertCss(cssString[, options]).
Error: insert-css: You need to provide a CSS string. Usage: insertCss(cssString[, options]).
    at insertCss (http://localhost:8080/build/index.html:12039:15)
    at http://localhost:8080/build/index.html:2341:9
    at Item.run (http://localhost:8080/build/index.html:6399:14)
    at drainQueue (http://localhost:8080/build/index.html:6369:42)
    at cleanUpNextTick (http://localhost:8080/build/index.html:6352:9)

fs is not defined
ReferenceError: fs is not defined
    at http://localhost:8080/build/index.html:43:12
    at http://localhost:8080/build/index.html:8938:19
    at Array.map (native)
    at Object.mcss (http://localhost:8080/build/index.html:8937:19)
    at http://localhost:8080/build/index.html:2793:22
    at Item.run (http://localhost:8080/build/index.html:6399:14)
    at drainQueue (http://localhost:8080/build/index.html:6369:42)

arj03:
@mixmix much better thanks 👍

PR #67, opened 2017-01-07
Network refactor

mixmix:
@ahdinosaur , @mmckegg would love a review on this before merging. 

Looks like this : 

selection_159

### DIFF
--- a/modules_extra/network.js
+++ b/modules_extra/network.js
@@ -3,13 +3,10 @@ const fs = require('fs')
 const h = require('../h')
 const human = require('human-time')
 
-const Struct = require('@mmckegg/mutant/struct')
-const Value = require('@mmckegg/mutant/value')
-const Dict = require('@mmckegg/mutant/dict')
-const toCollection = require('@mmckegg/mutant/dict-to-collection')
-const mutantMap = require('@mmckegg/mutant/map')
-const when = require('@mmckegg/mutant/when')
-const computed = require('@mmckegg/mutant/computed')
+const {
+  Struct, Value, Dict,
+  dictToCollection, map: mutantMap, when, computed
+} = require('@mmckegg/mutant')
 
 //var avatar = plugs.first(exports.avatar = [])
 //var sbot_gossip_peers = plugs.first(exports.sbot_gossip_peers = [])
@@ -216,7 +213,7 @@ function obs_gossip_peers (api) {
 
   refresh()
 
-  return toCollection.values(state)
+  return dictToCollection.values(state)
 
   function refresh () {
     api.sbot_gossip_peers((err, peers) => {
--- a/modules_extra/network.js
+++ b/modules_extra/network.js
@@ -1,5 +1,5 @@
 var isVisible = require('is-visible').isVisible
-var h = require('hyperscript')
+var h = require('../h')
 
 //var avatar = plugs.first(exports.avatar = [])
 //var sbot_gossip_peers = plugs.first(exports.sbot_gossip_peers = [])
@@ -105,13 +105,8 @@ function duration (s) {
 exports.create = function (api) {
 
   return {
-    menu_items: function () {
-      return h('a', {href: '#/network'}, '/network')
-    },
-
-    builtin_tabs: function () {
-      return ['/network']
-    },
+    menu_items: () => h('a', {href: '#/network'}, '/network'),
+    builtin_tabs: () => ['/network'],
 
     screen_view: function (path) {
 
@@ -136,9 +131,9 @@ exports.create = function (api) {
               || b.stateChange - a.stateChange
             )
           }).forEach(function (peer) {
-            ol.appendChild(h('div',
+            ol.appendChild(h('div', [
               api.avatar(peer.key, 'thumbnail'),
-              h('div',
+              h('div', [
                 peer.state || 'not connected',
                 ' ',
                 getType(peer),
@@ -150,16 +145,18 @@ exports.create = function (api) {
                 h('label',
                   {title: new Date(peer.stateChange).toString()},
                   peer.stateChange && ('(' + human(new Date(peer.stateChange))) + ')')
-                ),
-                'source:'+peer.source,
-                h('pre', legacyToMultiServer(peer)),
-                h('button', 'connect', {onclick: function () {
-                  api.sbot_gossip_connect(peer, function (err) {
+              ]),
+              'source:'+peer.source,
+              h('pre', legacyToMultiServer(peer)),
+              h('button', {
+                'ev-click': () => {
+                  api.sbot_gossip_connect(peer, (err) => {
                     if(err) console.error(err)
                     else console.log('connected to', peer)
                   })
-                }})
-              )
+                }},
+                'connect'
+              )])
             )
           })
 
--- a/modules_extra/network.js
+++ b/modules_extra/network.js
@@ -1,20 +1,23 @@
+const fs = require('fs')
 const isVisible = require('is-visible').isVisible
 const h = require('../h')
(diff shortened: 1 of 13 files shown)


dominictarr:
:+1:

mixmix:
Found this was not polling correctly. It's now polling every 5s (regardless of whether it's open).

Fix to mutant proposed to get this working here : https://github.com/mmckegg/mutant/issues/4

modules_extra/network.js
```diff
@@ -100,77 +95,166 @@ function duration (s) {
...
+      ])
+    ])
+  }
+}
 
-      })()
-
-      return h('div.column.scroll-y', ol)
-    }
+function obs_gossip_peers (api) {
```
ahdinosaur: ooh, is this a module?

mmckegg:
@mixmix mmckegg/mutant#4 has now been merged 😸

modules_extra/network.js
```diff
@@ -100,77 +95,166 @@ function duration (s) {
...
+    mcss: () => fs.readFileSync(__filename.replace(/js$/, 'mcss'), 'utf8')
+  }
+
+  function screen_view (path) {
+    if (path !== '/network') return
+
+    var peers = obs_gossip_peers(api)
+
+    return h('div', { style: {'overflow':'auto'}, className: 'column scroller' }, [
+      h('Network', [
```
ahdinosaur: omg this is so much better :clap:

modules_extra/network.js
```diff
@@ -100,77 +95,166 @@ function duration (s) {
...
+              api.avatar_image_link(key()),
+            ]),
+            h('section.name', [
+              api.avatar_name_link(key()),
+            ]),
+            h('section.type', [
+              computed(peer, getType),
+            ]),
+            h('section.source', [
+              'source: ',
```
ahdinosaur: should these labels ("source:", "state:", "rtt:", "skew:") be given a class so they can be styled?

modules_extra/network.js
```diff
@@ -100,77 +95,166 @@ function duration (s) {
...
 
-      })()
-
-      return h('div.column.scroll-y', ol)
-    }
+function obs_gossip_peers (api) {
+  var timer = null
+  var state = Dict({}, {
+    //
+    // TODO - get this api implemented in mutantDict
```
ahdinosaur: this is implemented now (https://github.com/mmckegg/mutant/issues/4), can we clean this up?

modules_extra/network.js
```diff
@@ -100,77 +95,166 @@ function duration (s) {
...
+          state.put(data.key, current)
+        } else {
+          current.set(data)
+        }
+      })
+    })
   }
 }
 
+function Peer () {
```
ahdinosaur: oh so nice :smile:

modules_extra/network.mcss
```diff
@@ -0,0 +1,113 @@
...
+    div.skew {
+      flex-basis: 8rem
+    }
+  }
+
+  section.address{
+    flex-basis: 100%
+
+    font-size: .8rem
+    $textSubtle
```
ahdinosaur: oooooooh yeah!

modules_extra/network.js
```diff
@@ -100,77 +95,166 @@ function duration (s) {
     return round(s)+'ms'
 }
 
+function peerListSort (a, b) {
```
mmckegg: Doesn't look like this sort function is being used yet.

modules_extra/network.js
mmckegg: I think it should be :)

modules_extra/network.js
```diff
@@ -100,77 +95,166 @@ function duration (s) {
...
+    screen_view,
+    mcss: () => fs.readFileSync(__filename.replace(/js$/, 'mcss'), 'utf8')
+  }
+
+  function screen_view (path) {
+    if (path !== '/network') return
+
+    var peers = obs_gossip_peers(api)
+
+    return h('div', { style: {'overflow':'auto'}, className: 'column scroller' }, [
```
mixmix: this is copied code - it notice the scroller className, which we've not got to yet.
mmmmm maybe put this off till later ?

modules_extra/network.js
mixmix: probably!

modules_extra/network.js
mixmix: hooray! yes will have a look at getting that sorted

modules_extra/network.js
mixmix: correct. I was wondering about using a filter instead of sort. I'm still pretty interested in making different sections to make this all easier to read / parse. 

What do you reckon @mmckegg

modules_extra/network.js
```diff
@@ -100,77 +95,161 @@ function duration (s) {
...
-
-      return h('div.column.scroll-y', ol)
+function obs_gossip_peers (api) {
+  var timer = null
+  var state = Dict({}, {
+    onListen: () => {
+      timer = setInterval(refresh, 5e3)
+    },
+    onUnlisten: () => {
+      clearInterval(timer)
```
mixmix: @mmckegg on Listen works not, but onUnlisten doesn't behave as expected - namely, when I either click off (styles are changed to `display: none`)  or close the tab (don't know what happens in background) the `/network` tab, the polling continues.

Not sure what onUnlisten is watching for, can you clarify

modules_extra/network.js
mixmix: i've put them in a label and styled them

modules_extra/network.js
mmckegg: It should be called whenever the elements that consume it are **removed** from the dom. So basically only when the tab is closed.

If this is not happening then it is probably a bug in `html-element`.

modules_extra/network.js
ahdinosaur: yeah okay sweet :cherries:

modules_extra/network.js
mixmix: woo! bugs. Can you reproduce what I've described - I just put a log in the `refresh`

mmckegg:
@mixmix I found the bug in mutant/html-element that was preventing the handlers from being cleaned up and have published a fixed version. I have gone ahead and bumped the mutant dep in this PR.

Also, [fixed another issue](https://github.com/ssbc/patchbay/pull/67/commits/1f965cd163e7c8df4b61d8f58aafb4b1bf15861b) where duplicate pubs were overriding each other. We were assuming that `peer.key` was enough to get a unique entry, but it turns out you also need to take in to account `host` and `port` otherwise the results are ambiguous.

mixmix:
woop!
great work @mmckegg 
Heads up that hypertabs might be changing to using `visibility: hidden` instead of `display: none`. Probs not relevant, but in case you've got mutant tricks for future lined up could be good to know

modules_extra/network.js
```diff
@@ -3,13 +3,10 @@ const fs = require('fs')
...
-const Struct = require('@mmckegg/mutant/struct')
-const Value = require('@mmckegg/mutant/value')
-const Dict = require('@mmckegg/mutant/dict')
-const toCollection = require('@mmckegg/mutant/dict-to-collection')
-const mutantMap = require('@mmckegg/mutant/map')
-const when = require('@mmckegg/mutant/when')
-const computed = require('@mmckegg/mutant/computed')
+const {
+  Struct, Value, Dict,
+  dictToCollection, map: mutantMap, when, computed
```
ahdinosaur: lowercase `map` is fine here

mixmix:
Yeah good point. Feel free to change it

PR #69, opened 2017-01-14
[WIP] ideal layout for nested message modules

mixmix:
this is a prototpye / talking piece for nested depject dev (and hopefully will exist in patchbay once the interface is more developed)


ahdinosaur:
thoughts: keen to move files into what i think are more descriptive categories for related modules. with `depject@3.1.0` we can now easily use [`bulk-require`](https://github.com/substack/bulk-require). i personally think `core`, `basic`, `extra`, ... should be top-level exports not directories.

mixmix:
yeah keen to talk about this in the weekly call (tmrw)

dominictarr:
does browserify understand bulk-require? I think it's probably better that we hand maintain those.

other than that, this looks fine to me. @evbogue what do you think?

mixmix:
lets talk about this on call - I think whether we use bulkify depends on
the module / folder strucure.

Apparently there's this - https://github.com/substack/bulkify

dominictarr:
oh yeah, instead of having an index.js contains nested objects it self (`avatar: {}, message: {}` etc) the message folder should have a index that provides the {} and then the top level should just be `message: require('./message')`.

and then, a client implementation can easily pick out those sub deps.

likewise, I reckon the way we do the minbay thing (with vertical tabs) is there is a module called `layout` which provides the top level wrapper (minbay, patchbay, and patchwork-next all have their own layouts) If we split the `basic` set into api and layout I think that will work. (and maybe we'll have to separate layout style from other style)

dominictarr:
@evbogue see my last comment

mixmix:
@dominictarr have made that change you suggested (`message: require('./message')` etc.)

I like your idea for a `layout` module. Have started an issue for it : https://github.com/ssbc/patchbay/issues/74

Testing works, though output is still pretty ugly. (a few `net.connect` errors)
It informed some changes to what is 'technically' in core according to deps (had to move some things in).

Things could work on (here or in another PR) : 
- [ ] testing output
- [ ] extending patchbay config to include a 'testing' env, which doesn't connect to your personal identity D:

mixmix:
Anyone recognise this error : 

```
TypeError: net.connect is not a function
    at Object.client (http://localhost:44827/bundle.js:107740:24)
    at Object.client (http://localhost:44827/bundle.js:107249:13)
    at Object.client (http://localhost:44827/bundle.js:107312:21)
    at module.exports (http://localhost:44827/bundle.js:136744:6)
    at http://localhost:44827/bundle.js:2506:7
    at tryConnect (http://localhost:44827/bundle.js:126121:11)
```

I've tried : 
- searching for anything like that in the project (nothing)
- browserifying a one of the test files and serching it, found only one instance  :
  ```js
  module.exports = function (opts) {
  opts.allowHalfOpen = opts.allowHalfOpen !== false
  return {
    name: 'net',
    server: function (onConnection) {
      var server = net.createServer(opts, function (stream) {
        onConnection(stream = toPull.duplex(stream))
      }).listen(opts.port)
      return function () {
        server.close()
      }
    },
    client: function (opts, cb) {
      var started = false
      var stream = net.connect(opts)  // <<<<<<<<<<<<<<<<<<<
        .on('connect', function () {
          if(started) return
          started = true
          cb(null, toPull.duplex(stream))
        })
        .on('error', function (err) {
          if(started) return
          started = true
          cb(err)
        })
    },
  ```

mixmix:
welp, I'm gonna merge this, because it's a stuctural change which is going to cause more problems the long it's open.

PR #72, opened 2017-01-16
Scroller module

mixmix:
@ahdinosaur would be interested in talking about this.

my motivations were : 
- extract Scroller as a module
  - add styling to scroller, so that Message has border ONLY when it's a child of a Scroller
- hide some of the specific pickiness of the scroller setup.

what I've made is on the edge of not worth it ... but then the scroller code everywhere is a bit weird.

I think what I've done is a new pattern in a couple of ways : 
- I have a weird 'content' keyword to note the position of the content div to be created
- I'm introduced a meta-template / constructor which other templates are using...

### DIFF
--- a/modules_basic/feed.js
+++ b/modules_basic/feed.js
@@ -5,6 +5,7 @@ var pull = require('pull-stream')
 var u = require('../util')
 
 exports.needs = {
+  build_scroller: 'first',
   sbot_user_feed: 'first',
   message_render: 'first',
   avatar_profile: 'first',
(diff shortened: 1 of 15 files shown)


mixmix:
@pietgeursen what to review this ?

dominictarr:
I should mention that since I wrote the scroller code I have thought that maybe it would be better to have a `state` object which was all the rows in the cache, instead of two streams.

mixmix:
can you expand ?
By two streams, you're talking about the pull streams from pull-scroll
which are appending/ prepending based on scroll right?

By rows, are you refering to you original column and row classes ?
By cache are you meaning that global CACHE?

modules_basic/scroller.js
```diff
@@ -0,0 +1,32 @@
+const fs = require('fs')
+const h = require('../h')
+
```
mixmix: @pietgeursen / @ahdinosaur this is the only file that really needs review

mixmix: then maybe look at one other file that implements the pattern. I'll post next to a good example

modules_basic/feed.js
```diff
@@ -25,42 +26,33 @@ exports.create = function (api) {
...
-      var content = h('div.column.scroller__content')
-      var div = h('div.column.scroller',
-        {style: {'overflow':'auto'}},
-        h('div.scroller__wrapper',
-          h('div', api.avatar_profile(id)),
-          content
-        )
-      )
+    const profile = h('div', api.avatar_profile(id))
+    var { container, content } = api.build_scroller({ prepend: profile })
```
mixmix: like here

dominictarr:
sorry!

so currently, you create two streams that forever expand the scroll window. eventually this is too much, especially if you are doing initial sync. so, one thing, is when you scroll down lots, it should start to _remove_ items from the top, but that means, when you scroll back up, it needs to create a new pull-stream, from the last point that you have. Instead of passing two streams, you'd need to pass two stream constructors.

that would give you a growing/shrinking array (or maybe an `{}` with number keys, so you can go negative if you scroll way up?) maybe this is would be useful as a state object for choo/inu/react? the part that doesn't map nicely onto those frameworks, however, is that you sometimes need to manipulate the current scroll position so that the screen doesn't jump around.

arj03:
In latest versions of patchbay changing tabs takes the tab jump to the top. Quite annoying for the main feed :) Does this fix that issue?

dominictarr:
@arj03 you are right! that must have been introduced in the change to `visibility: hidden`?

mixmix:
yeah I noticed that side effect too. Fixing bugs introducing bugs D:
This doesn't solve that.

mixmix:
Yeah that ideal Scroller sounds awesome @dominictarr - I've [opened an issue](https://github.com/dominictarr/pull-scroll/issues/6) on pull-scroll to track this

Given I think that's future dev, what do people think of this as an intermediate step?

mixmix:
@arj03 @dominictarr fix to scroll-position loss here : https://github.com/dominictarr/hypertabs/pull/5

modules_basic/scroller.js
```diff
@@ -0,0 +1,32 @@
...
+}
+
+exports.create = function (api) {
+  return {
+    build_scroller,
+    mcss: () => fs.readFileSync(__filename.replace(/js$/, 'mcss'), 'utf8')
+  }
+
+  function build_scroller ({ prepend = [], append = [] } = {}) {
+    var content = h('section.content')
```
pietgeursen: const

pietgeursen:
standard style?

modules_basic/scroller.js
mixmix: done

mixmix:
thanks for the review @pietgeursen 
I've squashed this and am merging

PR #73, opened 2017-01-20
Notify

dominictarr:
This PR makes tabs go red when there is new content in them.
switching to the tab removes the activity indicator.
we need better styling (red dot instead of whole tab being red?)
and to apply to tabs other than public,

and maybe to merge with https://github.com/ssbc/patchbay/pull/72 ?


util.js
```diff
@@ -48,4 +48,19 @@ exports.next = function (createStream, opts, property, range) {
 }
 
 
+exports.ScrollNotify = function (div, scroller) {
+  var cl = 'hypertabs--notify'
+  live.observ(function (len) {
```
mixmix: live is undefined @dominictarr

modules_basic/public.js
```diff
@@ -34,7 +34,7 @@ exports.create = function (api) {
 
         pull(
           u.next(api.sbot_log, {old: false, limit: 100}),
-          Scroller(div, content, api.message_render, true, false)
+          ScrollNotify(div, Scroller(div, content, api.message_render, true, false))
```
mixmix: In my code I changed `div` to be called `container`. I think the div you're giving to `ScrollNotify` should  not be the container, rather the hypertab tab, which are defined in `modules_core/tabs` right ?

dominictarr: yes it's the container.

mixmix:
still getting an error : 

```
Uncaught TypeError: scroller.observ is not a function

/home/mix/projects/SSBC/patchbay/util.js:53
TypeError: scroller.observ is not a function
    at exports.ScrollNotify (/home/mix/projects/SSBC/patchbay/util.js:53:12)
    at Object.screen_view (/home/mix/projects/SSBC/patchbay/modules_basic/public.js:37:11)
    at Object.screen_view (/home/mix/projects/SSBC/patchbay/node_modules/depject/apply.js:15:30)
    at /home/mix/projects/SSBC/patchbay/modules_core/tabs.js:67:20
    at Array.forEach (native)
    at Object.<anonymous> (/home/mix/projects/SSBC/patchbay/modules_core/tabs.js:66:11)
    at Object.screen_view (/home/mix/projects/SSBC/patchbay/node_modules/depject/apply.js:15:30)
    at Array.<anonymous> (/home/mix/projects/SSBC/patchbay/modules_core/app.js:27:22)
    at Object.<anonymous> (/home/mix/projects/SSBC/patchbay/index.js:9:9)
    at Module._compile (module.js:556:32)
```

dominictarr:
oh I should have mentioned that this depends on my PR to pull-scroll https://github.com/dominictarr/pull-scroll/pull/7

dominictarr:
oh I should have mentioned that this depends on my PR to pull-scroll

mixmix:
that makes more sense.
Have linked it in, _now we play the waiting game_

util.js
```diff
@@ -48,4 +48,19 @@ exports.next = function (createStream, opts, property, range) {
...
+  var cl = 'hypertabs--notify'
+  scroller.observ(function (len) {
+    if(div.classList.contains(cl) != (len != 0)) {
+      if(len) div.classList.add(cl)
+      else div.classList.remove(cl)
+    }
+  })
 
+  div.addEventListener('focus', function () {
+    live.visible()
```
mixmix: should be `scoroller.visible()`

mixmix:
ok you can stimulate it by publishing a private message ... which highlights the `/public` tab.

notes : 
- (aside) I don't think private messages should be rendered there.
- If i'm already at the top-limit of a feed, `pull-scroll` won't prepend.
  - current behaviour : I have to scroll down, then content gets prepended, very counterintuitive, and I'd not noticed it to the point of my way of using patchbay is to `ctrl+r` to get new content D:
  - I reckon something like in patchwork-next, which says 'show new messages would be cool. I think your `visible` method might help here. Maybe it should be called `go` or `add` because `visible` is presuming a particular use?

util.js
```diff
@@ -48,4 +48,19 @@ exports.next = function (createStream, opts, property, range) {
 }
 
 
+exports.ScrollNotify = function (div, scroller) {
+  var cl = 'hypertabs--notify'
+  scroller.observ(function (len) {
+    if(div.classList.contains(cl) != (len != 0)) {
+      if(len) div.classList.add(cl)
+      else div.classList.remove(cl)
+    }
```
mixmix: **[curiosity]**: I honestly can't see how this works, but can see it working in the browser.
Here `div` is the public scroller container. I can see `.hypertabs--notify` getting added to that. 

It's also getting added to the tab, class `.hypertabs__tab`.
What wizardry is this?

util.js
```diff
@@ -48,4 +48,19 @@ exports.next = function (createStream, opts, property, range) {
 }
 
 
+exports.ScrollNotify = function (div, scroller) {
+  var cl = 'hypertabs--notify'
+  scroller.observ(function (len) {
+    if(div.classList.contains(cl) != (len != 0)) {
+      if(len) div.classList.add(cl)
```
mixmix: while you're at it you could do something like `div.dataset.unreadCount = len` so we can get a counter up there.

mixmix:
@dominictarr I've wired this up to work with the new hypertabs.
I'm not sure  Ifeel great about how this plugs in, too tired to imagine alternative atm.

Things to know - the hypertabs interface is different, when you go `tabs.add(el)`, `el` is wrapped in a `div.page` before it's mounted. The mutation observer is set to watch this 'page'. 
I did this to make the box model more predicatable, but it's got the annoying side effect that you can't just class toggle the scroller, you have to toggle the page (which is the `scroller.parentNode`).

Maybe it feels a bit much like JQuery, where I would rather this felt like redux.
Maybe the the pattern of `build_scroller` should be that when you call it, it provides you with the (scroller) `container`, the `content` and an observable that the when modified triggers the Scroller to mark its page notify ...

util.js
mixmix: fixed

util.js
mixmix: what a jerk.
lets got minimal feature

PR #76, opened 2017-01-21
add Markdown mcss

mixmix:
Mainly added this to style the blockquote.
Feels like a nice clean easy way to add styles safely to markdown chunks.

Does ssb-markdown bring any styles in with it ?

### DIFF
--- a/modules_basic/markdown.js
+++ b/modules_basic/markdown.js
@@ -1,26 +1,25 @@
-var markdown = require('ssb-markdown')
-var h = require('hyperscript')
-var ref = require('ssb-ref')
+const renderer = require('ssb-markdown')
+const fs = require('fs')
+const h = require('../h')
+const ref = require('ssb-ref')
 
 exports.needs = {
   blob_url: 'first',
   emoji_url: 'first'
 }
 
-exports.gives = 'markdown'
+exports.gives = {
+  markdown: true,
+  mcss: true
+}
 
 exports.create = function (api) {
-
-  function renderEmoji(emoji) {
-    var url = api.emoji_url(emoji)
-    if (!url) return ':' + emoji + ':'
-    return '<img src="' + encodeURI(url) + '"'
-      + ' alt=":' + escape(emoji) + ':"'
-      + ' title=":' + escape(emoji) + ':"'
-      + ' class="emoji">'
+  return {
+    markdown,
+    mcss: () => fs.readFileSync(__filename.replace(/js$/, 'mcss'), 'utf8')
   }
 
-  return function (content) {
+  function markdown (content) {
     if('string' === typeof content)
       content = {text: content}
     //handle patchwork style mentions.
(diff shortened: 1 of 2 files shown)


modules_basic/markdown.js
```diff
@@ -1,26 +1,25 @@
...
-  function renderEmoji(emoji) {
-    var url = api.emoji_url(emoji)
-    if (!url) return ':' + emoji + ':'
-    return '<img src="' + encodeURI(url) + '"'
-      + ' alt=":' + escape(emoji) + ':"'
-      + ' title=":' + escape(emoji) + ':"'
-      + ' class="emoji">'
+  return {
+    markdown,
+    mcss: () => fs.readFileSync(__filename.replace(/js$/, 'mcss'), 'utf8')
```
dominictarr: this sort of thing breaks browserify, it can't eval that statically. this would be better as `path.join(__filename, 'markdown.mcss')`

dominictarr:
ssb-markdown doesn't currently require any special styling

modules_basic/markdown.js
mixmix: maybe I got you wrong, but I think https://www.npmjs.com/package/brfs has us convered here ?
I think it was specifically designed for this.

trying `brfs modules_basic/markdown.js` shows it evals fine.

clehner:
On Sat, 21 Jan 2017 14:35:42 -0800
mix irving <notifications@github.com> wrote:

> +    mcss: () => fs.readFileSync(__filename.replace(/js$/, 'mcss'), 'utf8')
> 
> maybe I got you wrong, but I think https://www.npmjs.com/package/brfs has us convered here ?
> I think it was specifically designed for this.
> 
> trying brfs modules_basic/markdown.js shows it evals fine.

from https://www.npmjs.com/package/brfs:

> Inside of each pathExpr, you can use statically analyzable expressions and these variables and functions:
> - __dirname
> - __filename
> - path if you var path = require('path') first
> - require.resolve()

So instead of `__filename.replace(/js$/, 'mcss')`
it should be `path.join(__dirname, 'markdown.mcss')`

mixmix:
Is npm run lite a sufficient test of this.
The dirname has been fine in all the other files to date.
Can change it just means adding path to all files which is a bit of a grind
if this currently works ok

modules_basic/markdown.js
dominictarr: @mixmix yes, but if you use a dynamic expression brfs can't understand it. what I recall, is that brfs can understand a very limited range of expressions, such as path.join(__filename, 'markdown.mcss') but not fs.readFileSync(__filename.replace(/js$/, 'mcss'), 'utf8')

dominictarr: oh, it uses https://www.npmjs.com/package/static-eval and maybe that can handle your expression now?

dominictarr: okay, sorry I was wrong, static-eval can handle a lot more than when I first used brfs.

PR #77, opened 2017-01-22
extract mcss module builder

mixmix:


mixmix:
yeah this definitely doesn't browserify down.
In contrast [this PR](https://github.com/ssbc/patchbay/pull/76) collapses down fine 

closing this

PR #78, opened 2017-01-23
Refactor profile

mixmix:
A newly modularised Profile Page

selection_171

features not pictured: 
- a description field (top right)
- followers, following (with smaller icons) 
- 

to develop still
- [ ] editable description field
- [ ] description field query / module (does exist?)
- [x] name aliases query (does exist?)
- [x] clear `Relationship status` (or similar) title
- [x] drain for streaming friends (so don't have to wait for the jarring collect)
- [ ] collapsed, expandable followers ... 
- [ ] clearer crop shading for hypercrop, also invert shadow if possible

### DIFF
--- a/modules_basic/avatar/profile.js
+++ b/modules_basic/avatar/profile.js
@@ -1,5 +1,8 @@
-var h = require('hyperscript')
-var pull = require('pull-stream')
+const fs = require('fs')
+const h = require('../../h')
+const pull = require('pull-stream')
+const { unique, collect, drain } = pull
+
 
 exports.needs = {
   avatar_image_link: 'first',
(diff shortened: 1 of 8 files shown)


dominictarr:
looks great! what do the numbers next to the names mean? that is number of feeds that use that name? It would be great to mouse over and get a popup on who uses that name.

---
I like that friends are big (because friends are important) but does this take up too much space?) Maybe there should be an inline tabs for that - though, prehaps this is a discussion for later, and we should merge this as is - 

- if there where tabs on the profile page, then there could be a "private" tab for messages between you and them. This would be really useful because it is quite hard to find an old private thread with a particular person!

modules_basic/avatar/profile.js
```diff
@@ -9,71 +16,71 @@ exports.needs = {
...
+        peer => rawFollows.push(peer), 
+        data => console.log('follows drain done', data)
       )
     )
+    pull(
+      api.followers(id),
+      unique(), 
+      drain(
+        peer => rawFollowers.push(peer), 
+        data => console.log('followers drain done', data)
```
mixmix: @dominictarr I've surfaced a bug with this - the pull stream is breaking - checking out the console log I've put in here catching some bitnation stuff.

dominictarr: uh, can you be more specific about the what sort of "break" this is?

dominictarr: do you mean you are not expecting followers stream to end?

dominictarr: it ends because it's not a live stream, https://github.com/ssbc/patchbay/blob/a2c499225bd1fcef80f7b99eeb8b826ab2814d0c/modules_basic/relationships.js#L37-L42

btw, on 58, it's a proper callback, so the signature is `function (err) {...}` not `data`

mixmix: yeah sorry that was a vauge message - was running to get out the door. 

What I'm seeing is this logged  :
```js
{message: "could not find matching timestamp for index:{"valu…25519","sequence":246},"timestamp":1481442875351}", name: "Error", stack: "Error: could not find matching timestamp for index…ttlebot/node_modules/level-sublevel/nut.js:121:19"}
```

I think it's a preature abort of the pull stream (I've read `pull.drain` a bit but some of the switches  I don't quite follow). 

I've looked closely at friendships and followers , and the 'status'  and think I might've seen it say "you are friends" but not seen myself in the friendship list.

Probably useful to understand what pull.drain is doing, and how best to handle that error - without the callback, it's breaking the interface....

dominictarr: okay I nearly couldn't find this thread again because github "reviews" are too complicated now. (it's this explicit "approve" thing, which is probably intended for enterprise) anyway,

that error is not about drain, it's from ssb-query. try deleting ~/.ssb/query and restarting sbot. I want to redo how this stuff works for websbot `%yYXsOihj80vH7fqL4HQiwdAHs594U8bPhCph5OCpCWs=.sha256` so I don't think we should worry about it too much right now.

mixmix:
@dominictarr correct number = number of uses of particular alias. I don't know if resiliant to a particular individdual spamming the publishing of a particular alias.
I agree the thing where you can see who declared that alias would be good but figure that can be added later.

Similar with inline tabs - they're a good idea, and lets add as an extension if needed / wanted. Turns out it's not a bit nicer - I was formerlly  showing all friends under both follows AND followers. Removing lightens the page significantly.

I think if this code looks alright merging it would be good.

---

After this is in, Id love some help writing a `about.description` query. I think this is contingent on some more design - as in which descriptions can we display, and how do people add/ edit them...

dominictarr:
merged into 6.4.0

PR #79, opened 2017-01-26
extract suggest-box into a helper module which builds suggest boxes

mixmix:
~~extract suggest logic into `built_suggest_box` and add some sweet styling~~

[EDIT]

**Context** :

We're using a module called `suggest-box`, which is responsible for the text-complete suggestions on a few types of mentions in the compose + search inputs.

This pull request : 
- adds images to @-mention and emoji-mention type suggestions
- takes advantage of suggest-boxes small text feature to space out the @-mention info
- extracts the suggest-box specific styling into a tighter (specificity) and clearer css module
- extracts a generic `suggest_box_builder`

### DIFF
--- a/modules_basic/avatar/image.js
+++ b/modules_basic/avatar/image.js
@@ -79,11 +79,6 @@ exports.create = function (api) {
     avatar_image_src: function (author) {
       var src = ready && avatars[author] ? api.blob_url(avatars[author].image) : genSrc(author)
 
-      if(!ready)
-        waiting.push(() => {
-          if(avatars[author]) src = api.blob_url(avatars[author].image)
-        })
-
       return src
 
       function genSrc (id) {
--- a/modules_basic/avatar/image.js
+++ b/modules_basic/avatar/image.js
@@ -77,9 +77,9 @@ exports.create = function (api) {
     },
 
     avatar_image_src: function (author) {
-      var src = ready && avatars[author] ? api.blob_url(avatars[author].image) : genSrc(author)
-
-      return src
+      return ready && avatars[author]
+        ? api.blob_url(avatars[author].image)
+        : genSrc(author)
 
       function genSrc (id) {
         if(cache[id]) return cache[id]
--- a/modules_basic/avatar/image.js
+++ b/modules_basic/avatar/image.js
@@ -12,7 +12,9 @@ exports.needs = {
 }
 
 exports.gives = {
-  connection_status: true, avatar_image: true
+  connection_status: true, 
+  avatar_image_src: true,
+  avatar_image: true
 }
 
 var ready = false
@@ -72,6 +74,24 @@ exports.create = function (api) {
 
         })
       )
+    },
+
+    avatar_image_src: function (author) {
+      var src = ready && avatars[author] ? api.blob_url(avatars[author].image) : genSrc(author)
+
+      if(!ready)
+        waiting.push(() => {
+          if(avatars[author]) src = api.blob_url(avatars[author].image)
+        })
+
+      return src
+
+      function genSrc (id) {
+        if(cache[id]) return cache[id]
+        var { src } = visualize(new Buffer(author.substring(1), 'base64'), 256)
+        cache[id] = src
+        return src
+      }
     },
 
     avatar_image: function (author, classes) {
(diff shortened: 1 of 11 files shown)


modules_core/search-box.js
```diff
@@ -14,17 +12,15 @@ exports.create = function (api) {
...
     var search = h('input.searchprompt', {
       type: 'search',
       placeholder: 'Commands',
-      onkeydown: function (ev) {
+      onkeydown: ev => {
         switch (ev.keyCode) {
           case 13: // enter
-            if (suggestBox && suggestBox.active) {
-              suggestBox.complete()
-              ev.stopPropagation()
```
mixmix: this check seems not to be needed

modules_core/search-box.js
```diff
@@ -47,19 +43,7 @@ exports.create = function (api) {
       }
     }
 
-    // delay until the element has a parent
-    setTimeout(function () {
```
mixmix: I'm not sure if it's terrible, but I just gave suggest-box what it wanted in the `build_suggest_box` function - a parent to reference.... 

It means this Timeout isn't needed ...

modules_core/suggest-mentions.js
```diff
@@ -24,14 +25,18 @@ exports.create = function (api) {
     return function (cb) {
       if(!/^[%&@]\w/.test(word)) return cb()
```
mixmix: hmmm, do we really want the to be showing images for `%` projects ? 
also `&` blobs lookup don't seem to be working

modules_core/suggest-mentions.js
```diff
@@ -40,18 +45,36 @@ exports.create = function (api) {
...
+              title: name,
+              subtitle: `(${rank}) ${id.substring(0,10)}`,
+              value: '['+name+']('+id+')',
+              rank,
+              image: api.avatar_image_src(id),
+              showBoth: true
             }
           }))
         })
+      } else if (/^%\w/.test(query)) {
```
mixmix: bit of duplication across these - might make sense to make a bunch of functions in here to give to different modules.

mixmix:
@dominictarr ! a juicy treat !

modules_basic/avatar/image.js
```diff
@@ -74,6 +76,24 @@ exports.create = function (api) {
       )
     },
 
+    avatar_image_src: function (author) {
+      var src = ready && avatars[author] ? api.blob_url(avatars[author].image) : genSrc(author)
+
```
mixmix: I'm not really sure if this pattern works (stollen from function below `avatar_image`). I.e. can you return a straight var early then modify it later?

dominictarr: no this won't work (if I understand what you are trying to do...) because you are gonna assign a image's src property to this right? remember the variable `src` is just a reference so when you do `image.src = src` then later set `src = src2` `image.src` will still have the old value.

dominictarr:
can you add more context to your description of this pr - I think this was mostly code which @clehner touched last so I'm not sure I remember exactly what it was doing

modules_basic/avatar/image.js
mixmix: yeah that's what I thought. 
The interface still seems to work despite this, so I'll jsut drop this

mixmix:
Yeah sorry I've been doing a bunch of late night coding and haven't been leaving very helpful messages.

**Context** :

We're using a module called `suggest-box`, which is responsible for the text-complete suggestions on a few types of mentions in the compose + search inputs.

This pull request : 
- adds images to @-mention and emoji-mention type suggestions
- takes advantage of suggest-boxes small text feature to space out the @-mention info
- extracts the suggest-box specific styling into a tighter (specificity) and clearer css module
- extracts a generic `suggest-box` module which takes care of a suggest-box peculiarity (requirement of a parentNode for a field you're still building), and dries up some repeated async code to help offer a cleaner interface.

mixmix:
@dominictarr is there any feedback you want to give / changes you'd like made?
If you don't have time, let me know and I'll source someone to review

modules_core/search-box.js
mixmix: fixed this with `on-load`

mixmix:
Published in 6.6.0

PR #81, opened 2017-01-29
bump scuttlebot version

mixmix:
I notice patchbay uses a version of scuttlebot that is behind by a major version. Not sure if there's good reason, but here's a conversation starter.

I was wondering about this after looking at some bugs with my profile feed.
**Should my sbot and patchbay be using the same sbot version ?**

my intuition from basic understanding of muxrpc is yes ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "pull-reconnect": "^0.0.3",
     "pull-scroll": "^1.0.1",
     "pull-stream": "^3.4.5",
-    "scuttlebot": "^8.7.2",
+    "scuttlebot": "^9.4.3",
     "setimmediate": "^1.0.5",
     "simple-mime": "^0.1.0",
     "split-buffer": "^1.0.0",


dominictarr:
hmm, I think the sbot in patchbay is only used when running embedded, so it doesn't actualy matter if you use it as a client.

mixmix:
oh interesting. yeah I saw the embedded stuff and it seemed like a little fork / experiment which might not be fully wired in - I didn't know so just left it.

Feels like there might be a couple of use cases it would be good to split out. Benefits would be that it would tidy up the code base, and force us to answer some important questions about how we're doing modules.

dominictarr:
I think we should have a bundled version that takes runs sbot for you so it can be easily bundled,
and see also %JDTx0jRXX8HX4E0QLyM4fsZUdvsSk8QkYRDLSbGicW0=.sha256 also, websbot will need to run in a webworker, which is also similar.

PR #82, opened 2017-02-02
Adds ctrl+scroll zooming in electron.

pietgeursen:


pietgeursen:
Dunno is peeps want this but I use it all the time in the browser and was missing it. I didn't even think to look in the menu where you can adjust the zoom. I just assumed it couldn't change. Derp. 
Thoughts @mixmix?

modules_core/tabs.js
```diff
@@ -191,6 +191,17 @@ exports.create = function (api) {
...
     if (process.versions.electron) {
+      var {webFrame} = require('electron')
+      
+      window.addEventListener('mousewheel', function (ev) {
+        var {ctrlKey, deltaY} = ev
+        if(ctrlKey){
+          var direction = (deltaY / Math.abs(deltaY))
+          var currentZoom = webFrame.getZoomLevel()
+          webFrame.setZoomLevel(currentZoom - direction)
+        }
```
mixmix: this looks great. I'll smoke test then merge.

I'm gonna change some of this to be styled a little more consisten with what I've been writing with es6.

dominictarr:
if I zoom right in using this, everything disappears and then to get it back I need to refresh (ctrl-r)
however, that doesn't happen when using the keyboard shortcut to zoom in (weird!)

mixmix:
maybe we could slow down the scroll zoo
I don't use this feature so happy to leave how it is untill a point where
it generates more need for change

PR #85, opened 2017-02-04
Add support for a channel input box

brhue:
Hey everyone!
This is a pull request to add a channel input to the message compose area. 
I threw this together because as a new user to patchbay it wasn't exactly clear to me how to make a post to the channels that I was seeing next to messages in the public feed.
While testing and exploring the client more I realized that being on the tab for a channel and making a post would create the post in that specific section, however I feel like this adds some clarity.

While on the /public feed the channel input is optional and while on a channel feed the input will be auto filled with the correct channel. The input also supports the suggest_box ui that shows up for mentions and on the search box.

If this is something you guys want to merge into main that'd be awesome, if not that's cool too and I'd be open for other ways I can help the project out.

### DIFF
--- a/modules_basic/compose.js
+++ b/modules_basic/compose.js
@@ -50,7 +50,18 @@ exports.create = function (api) {
 
     var channelInput = h('input.channel', {
       placeholder: '#channel',
-      value: meta.channel || ''
+      value: meta.channel ? `#${meta.channel}` : '',
+      disabled: meta.channel ? true : false
+    })
+
+    channelInput.addEventListener('keydown', function (e) {
+      console.log(e)
+      if (this.value.startsWith('#') && this.value.length === 1) {
+        this.value = ''
+        return
+      }
+      if (this.value.startsWith('#')) return
+      this.value = `#${this.value}`
     })
 
     if(opts.shrink !== false) {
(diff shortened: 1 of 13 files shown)


dominictarr:
This looks good!

two small things I noticed when testing this:

If you select the channel suggester without entering any text into the textarea, then it collaspse the composer. selecting the channel suggester should keep it open.

Also, you still have to type the "#" to get a channel. maybe there is a simple way to always force the `#` sigil to be entered?

mixmix:
Looks great @brhue , I've been wanting this for a while

To summarize @dominictarr 's point as user stories (and add one more) : 
- [ ] when I click compose, then click channel, I expect the compose to stay expanded
- [ ] when I type a channel I expect the # symbol to be there to teach me about prefixes through consistent display
- [ ] when I reply to a message that had a channel, channel selection is pre-populated (and maybe even unchangeable, up for discussion)

modules_basic/like.js
```diff
@@ -43,8 +43,8 @@ exports.create = function (api) {
     if (votes.length === 0) return
 
     digs.textContent = votes.length > 4
-      ? votes.length + ' 🗸'
-      : Array(votes.length).fill('🗸').join('')
+      ? votes.length + ' \u2713'
+      : Array(votes.length).fill('\u2713').join('')
```
mixmix: This could have been a different pull request - only just saw it.
I've fixed and published a solution alread !

brhue:
Awesome! Thanks for the feedback guys. Gonna start working on these suggestions.

modules_basic/compose.js
```diff
@@ -47,6 +48,22 @@ exports.create = function (api) {
       placeholder: opts.placeholder || 'Write a message'
     })
 
+    var channelInput = h('input.channel', {
+      placeholder: '#channel',
+      value: meta.channel ? `#${meta.channel}` : '',
+      disabled: meta.channel ? true : false
+    })
+
+    channelInput.addEventListener('keydown', function (e) {
```
mixmix: i'm going to bring some of this into style alignment (I've been moving more anonymous functions like this to fat arrow functions), and remove the console.log.

Happy to merge this and make that change.

awesome work ! :taco:

dominictarr:
@brhue thanks for the PR great work!

PR #89, opened 2017-02-08
Module restructure

mixmix:
This is a large change. Looking at the diff won't help I don't think.
I recommend you pull the branch down and have a look at it.

Two large patterns : 
1. put modules in folders which match their message type
  - except `pages` (for full page views) and `helpers` (which are message agnostic)
  - the modules_core folder is a little different and might follow a different pattern?
2. nested module export/ import
  - this actually helped me see clearer domains, and suggests clearer naming.

I'm a little apprehensive because this took a few hours, and I want to make sure I'm not badly baised by that investment. I think it feels like a good pattern. 
Thoughts @dominictarr @ahdinosaur  ?


mixmix:
@evbogue in terms of importing I'm hoping this will make life easier to grab the chunks you might want from patchbay.
While there are some significant name changes, the intention is to make it easier for people to learn about message types, and to think about what they're building / adding / requiring.

At the moment this branch doesn't compile down to lite client / bundle - this is definitely something any final version will need

index.js
```diff
@@ -1,10 +1,20 @@
...
-  require('./modules_extra'),
-  require('./modules_basic'),
-  require('./modules_core')
+  load('modules_extra'),
+  load('modules_basic'),
+  load('modules_core')
 ).app[0]()
 
+
+function load (moduleSetName) {
```
ahdinosaur: this breaks `browserify`-ability

modules_extra/notifications.js
```diff
@@ -106,15 +110,15 @@ exports.create = function (api) {
   }
 
   function getFirstMessage(feedId, cb) {
-    api.sbot_user_feed({id: feedId, gte: 0, limit: 1})(null, cb)
+    api.sbot.user_feed({id: feedId, gte: 0, limit: 1})(null, cb)
```
ahdinosaur: now that we have proper namespacing through nesting, can we use `camelCase`?

@dominictarr curious, what was the initial reason for `snake_case` in patchbay?

modules_extra/search.js
```diff
@@ -7,14 +7,20 @@ const Scroller = require('pull-scroll')
...
+  },
+  sbot: {
+    log: 'first',
+    fulltext_search: 'first'
+  }
 }
 
 exports.gives = {
-  screen_view: true,
+  page: true,
```
ahdinosaur: yus!

test/modules.js
```diff
@@ -0,0 +1,38 @@
...
+process.env.ssb_appname = 'test'
+
+console.log(path.join( __dirname, '../modules_core'))
+
+
+console.log(`
+/////////
+// CURRENTLY TEST BROKEN
+//
+// It's bulk-require. I've tried using bulkify but it it does not seem to play well with brfs
```
ahdinosaur: as i said before, `bulkify` and `brfs` play fine together. what order are they being run in? we want `bulkify` to be first since it adds new modules. but looking at the package.json you don't have `bulkify` at all there.

index.js
mixmix: yeah I know this is a convenience method.
I spend an hour on brfs + bulkify playing well together and did all the things down to the point of hard-coding addresses (removing alllllllllll fanciness). NOPE.

I figured I would loop back around once there was more clarity about this approach. Worst case can re-introduce the `index.js` files hand-written

modules_extra/notifications.js
mixmix: I don't know, I thought that was you O.o

mixmix: I'm actually fine with either. I have found that using snake case makes the depject modules stand out loud and clear, which has been a god-send for moving and renaming things

index.js
ahdinosaur: i'm happy to help get `brfs` and `bulkify` working well together.

test/modules.js
ahdinosaur: this is broken because `modules_core/tabs.js` line 5. it's trying to browserify `require('electron')`, which is failing.

ahdinosaur: @mixmix here's a quick fix: https://github.com/ahdinosaur/patchbay/commit/e48467e995134f41edd3d38570e7d4dff82e3f54. the tests still fail, but for a different reason. :wink:

mixmix:
hey @mmckegg I see your cheeky commit (thanks!) and that you branched this to explore that mcss alternative.
Do you have any comments on this PR as a patch* dev ?

mixmix:
Unless there's major objection or some further conversation, I plan to progress to mergeing this at the end of this weekend (NZT). 
Please let me know if this is a bad idea and I'll make more space :horse: 

---

It's the sort of PR that is costly to keep up to date - I notice I've already tainted this branch by adding  ttps://github.com/ssbc/patchbay/pull/89/commits/5ea4a58d2df21ab1dbd78a52f394a558f8ba491a which adds profile to the menu

cc @dominictarr @ahdinosaur

dominictarr:
I'm against using bulk-require. I think it's much better to be explicit, and it preserves browserify, which means people who would rather use their client via the browser than electron still have a working client. It may in theory be possible to make bulk-require work with browserify, but you have to use some clever thing, and it's better to just be explicit and not require a clever thing.

Okay, it's a pain when you restructure everything, but you just have a script that can regenerate the index.js for that situation... There is one in `./scripts/dir.js`

I'm also suspucious of directories with only one file in them. e.g. `./modules_core/about`

I totally agree that patchbay could be better organized, but am not sure about message type being the separator. Also, there are some places where there is a directory that doesn't map to a message type. for example `module_basic/message` and `module_basic/pages`. In my opinion, folders should represent a _distinction_. If files are in a folder, it should mean there is something particular about those files, that they share. It's like chapters in a non-fiction book. (for this reason, I hate "lib" folders, because it doesn't really mean anything, but I am okay with "test", and "examples". [pull-stream](https://www.html5rocks.com/en/tutorials/file/filesystem/) has `sinks`, `sources`, and `throughs` directories (which separate streams by type). The same structure is repeated in the `docs` folder.

I don't have a good answer for what the first organizing principle should be. maybe it should be things that only provide apis, renderers, and generic widgets?

modules_extra/notifications.js
dominictarr: Yeah, it was something to do with not having nesting in depject 1. I never used snake case until I started using sodium. I kinda liked it there, and it showed that anything with snake case was close to the metal (c!) so, to be taken seriously. By the time it could be rewritten to be camel, when I was refactoring for depject2, I left it out because I just wanted to get it working, it's something that is actually quite a bit of work to change, but little real benefit.

I generally try to avoid camel names, unless there just isn't a horse name that suffices. (unless, like with directories, the prefix implies a distinction `isTest` (pure, returns boolean) `createThing` (returns a a thing with state)

dominictarr:
prehaps a better question, what is the hard part to keep up to date? is there something that could change to make this sort of reorginization easier?

dominictarr:
ps, my apologies for not getting to this sooner.

dominictarr:
the github view hides comments from commits on lines which are no longer on top (which is quite confusing, because you still see that in the email view) anyway I noticed there where quite a few comments about wrestling with browserify/bulk-require

mixmix:
the lite client build works - have just confirmed this. Mikey and I resolved the bulk require. I personally think having it makes development easier, but I didn't know about your auto-script. Open to checking that out - I think this is a fairly minor issue we can resolve after we've talked more about the split

the update was a pain to keep up to date because this isn't just moving files to new directories - in a couple of cases, the responsibility of rendering was kinda spread across a couple of domains. So for examples the page : `profile` is now a lot nicer, because I split `about`type components (name, avatar) from `contact` type (friends, follows, follow/unfollow).

---

The thinking of splitting is perhaps more accurately described as splitting over domains - I see a bunch of about related stuff - from raw rendering and serving up of common avatar components to some higher-order stuff like `signifiers` (which is an abstraction over many abouts).
I think pages are another type of abstraction - a synthesis of several domains (e.g. profile which is about + contact + post)

## things I've enjoyed about this pattern so far

I thought I wrote these down, but maybe it was in another context

The division provided
- teaches new users about message types
- helps users think clearly about the domain they are calling
- the modules_basic folder is now emmanently navigable (to someone who is not the author), as you don't have parse the semantics of a filename to figure out if it's in the area you're trying to change.
- it has suggested clearer naming patterns (render.js contains rendering a message type / domain)
- nested modules encourages clear thinking about the domain you're adding to, and makes naming easier
- shorter, clearer file names

I'm not sure helpers is quite right, but it feels better to have some of the more raw decoration type modules not muddled in with eveything else

## things i've not enjoyed

- `modules_core` might not follow such a clear pattern
- `modules _extra` doesn't have as clear message types, might suggest looser domains like 'git' or 'music'

---

I'm up for trying some different splits. I think moving files into a different tree and seeing what that looks like might be a better idea than trying to wire it up so it works (that  can take quite a while).

I'm pretty keen for the domain 'pages' to persist (regardless of name). I think api-like stuff could be moved into its own folder. I'm interested in how to drop folders without returning to stacks of verbose naming like `message_link.js`.

ahdinosaur:
> It may in theory be possible to make bulk-require work with browserify, but you have to use some clever thing

this clever thing exists: [`substack/bulkify`](https://github.com/substack/bulkify)

> I don't have a good answer for what the first organizing principle should be. maybe it should be things that only provide apis, renderers, and generic widgets?

hm. for reference, what might be inspiring these changes: [my preferred way of organizing code into directories](https://github.com/enspiral-root-systems/catstack/blob/914d1a7c6730bef124f14ab3acb814cce0cf8fe4/README.md#domain-overview), my inspiration from [this part of this talk](https://youtu.be/WpkDN78P884?t=7m).

in this way, there are two layers of organization: 1) what is the module's "domain" (grouping based on conceptual cohesion), 2) what is the module's "type" (grouping based on how module interfaces with other modules).

if we apply this here, the first layer is really anything, the second layer is how we categorize possible exports: pages (screen views), message content renders, observables, elements / components / widgets, apis, what else?

but at the moment we our first layer is core, basic, extra, which means we either have 3 levels of directories or that is the first layer (which i'm not sure fits my idea of conceptual cohesion). i'd be keen to drop these directories and move extra code to a separate repo, but that's probably another discussion.

mixmix:
+1 a module directory and extraction to other repos - I think this will put us on a good path to optimising for modularity and remixing. I guess the question is when is the time to do that. It feels like core + basic are actually pretty fundamental. trying to manage those not being interdependant has been grindy. mmm Needs to happen soon so we don't get inseperable spaghetti mess...

dominictarr:
> Mikey and I resolved the bulk require.
Okay, I just don't want to merge anything that breaks lite client, because that is a hard breaking change for some users. Additional Context: I don't like doing "clever" things with browserify because I of plans I have about merging depject layers, and non-simple browserify setups will complicate that. Also, note: an explicit import makes it easy to disable specific modules during dev.

> the update was a pain to keep up to date because this isn't just moving files to new directories - in a couple of cases, the responsibility of rendering was kinda spread across a couple of domains.
 So for example the page : profile is now a lot nicer, because I split abouttype components (name, avatar) from contact type (friends, follows, follow/unfollow).

If this PR was just that, I'd be 👍 merge immediately! I'm willing to consider merging to get that in, as long as we leave the discussion about how best to organize the folders remains open.

@ahdinosaur for you, what is a "domain". This part I don't really understand.

One idea, that I orginally had but didn't get around to, is that you should be able to run _just_ core, or just core+basic and still get a functioning app. Except having just core would be something super minimal, like no threads, and maybe just JSON rendering.

mixmix:
I think the runnable core idea might be possible, but there are actually quite a few polluting cross-links into basic at the moment. I think there needs to be a seperate attempt at defining a 'minimal'

I'm open to refactoring the groupings in the future.
I think while it's uncomfortable holding this topic / PR open, I think we should take the discussion to '80%' not 50%, because the opinions here will inform future work

mixmix:
@pietgeursen and I are jamming on this on a big piece of paper if anyone wants to join (physically or remote).

questions : 
- what is meant by API  (is it like a helper function, like `mardown`, `blob_url`, `timestamp` ?)
- what is a widget? (is it like a partial, an element of interaction, like scroller, reply, ...)

mmckegg:
@mixmix I have some _opinions_, but I would prefer to discuss in person. Will you be around (available) sometime Monday afternoon? Me and @ahdinosaur have talked about peering on some ssb stuff then.

mixmix:
I think I have to work then D: Want to hang out today @mmckegg @ahdinosaur ?

dominictarr:
A widget is not a partial. A widget is something that is not tied to a specific application - it's a hinge that can be used on multiple completely different applications. it's the "hinge" in the IKEA architecture. pull-scroll, hypertabs and suggest-box are widgets. The composer is pretty much tied to ssb apps, so that is a partial. Another sign: if it works as an npm module then it's probably a widget. An application may have it's own wrapper around a widget, but that wrapper will be "thin" i.e. not much complex code.

Partials are only reused within one application, not much is gained from publishing them (that is, it's unlikely another app will depend on it) this is the problem with https://github.com/patchkit

all the `sbot_*` stuff is definitely api, also stuff that wraps that, like `relationships.js` (which was moved in this pr, i think)

mmckegg:
@mixmix yeah, sure. Where you at?

ahdinosaur:
> @ahdinosaur for you, what is a "domain". This part I don't really understand.

to me, a "domain" is a rough grouping of modules that you feel are related and therefore want to be cohesive inside and decoupled outside. it's very sentimental and subjective, as opposed to the next layer of grouping which is more objective based on the "type" / "shape" of export.

examples:
- [%holodex](https://github.com/holodex/app/tree/8f91a1c40ae9baf7a6645507b3f73127d4a5aa5f/dex): each directory is a domain.
-  [%patchcore](https://github.com/ssbc/patchcore): the repo is a domain.

ahdinosaur:
> %patchcore: the repo is a domain.

ignore this, it's many domains.

PR #90, opened 2017-02-10
remove scoping on mutant dependency

mmckegg:
Managed to scare of the squatter and nabbed [unprefixed `mutant`](https://www.npmjs.com/package/mutant). 

**This PR updates patchbay to use this version instead!**

### DIFF
--- a/h.js
+++ b/h.js
@@ -1,4 +1,4 @@
-const { h } = require('@mmckegg/mutant')
+const { h } = require('mutant')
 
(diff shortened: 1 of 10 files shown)


mixmix:
ah damn, I just changed it over an hour ago and hadn't seen this D:
Thanks for the PR though. 
Got the new micro-css with clearer styles in too.... oh that reminds me of a change I should make to hypertabs docs ... sigh

Issue #86, opened 2017-02-04
overridable styles


dominictarr:
I want to think about how styles work in patchbay. The current way isn't working for everyone, and this is causing @evbogue to maintain a separate fork of patchbay (that isn't efficient!). ev is the vanguard of this, but there are also other projects based on tweaking patchbay (entropy, pangea) so I think this is an important thing for us to consider.

The first thing that comes to mind is that patchbay has a `mcss` socket which is called as a map (calls all `mcss` plugs), but it also has a `css` socket!

My first suggestion is to use only `css` and just have plugs which want to use `mcss` compile it themselves, and give a `css` plug only https://github.com/ssbc/patchbay/blob/master/modules_core/style/styles.js#L24 This would mean there is only one thing to think about.

The problem with calling things as a map is that there is no way to remove items. one simple solution that comes to mind immediately is to instead use a `reduce` socket. instead of just returning the css for the current module, you'd have a signature like this:

``` js
var my_style = '' //your css here, generated from mcss or whatever
module.exports = {
  gives: {css: true},
  create: function () {
    return {css: function (style) { return style + my_style }}
  }
}
```
that would mean you could have a module that just ignored every other `css` plug like this:

``` js
//no-style.js
module.exports = {
  gives: 'css',
  create: function () {
    return function (style) { return '' } //set style to empty
  }
}
```

then you'd load that in when you combine the depject modules...

``` js
require('depject')(
  require('patchbay'),
  require('./no-style'),
  require('./modules-with-new-styles')
)
```

That is one approach that would require minimal changes to the current code. I think there are a few other approaches that could work as well, and we should explore these a bit.

dominictarr:
oh I just realized that @ahdinosaur already suggested reduce in this thread. http://viewer.scuttlebot.io/%25xH+N+IEwbMAiKYbmk38S7pKbX3NoQqmNaw27eqYk85s=.sha256#%25G54%2BFfturihfEtcIZ3L5oC5Dk1FMHOJCXc7iCYvGTAA%3D.sha256

evbogue:
Returning to an easy theme-able Patchbay would be amazing.

How do we proceed in making this happen?

ahdinosaur:
> an easy theme-able Patchbay

@evbogue what does this mean to you? as in, can you describe your ideal way to theme Patchbay? i'd love to help "make it happen" but i'm not sure what is wrong with our current approach and what we want instead.

dominictarr:
[...]

PR #91, opened 2017-02-10
[proposal] Extract styles from depject

mmckegg:
I've been having some ideas about styling spawned from [recent discussion](https://github.com/ssbc/patchbay/issues/86). Rather than ramble on about them, I thought I'd just go ahead a try it!

I've **extracted all of the styling out of depject** (moving it to styles/*) and made it the responsibility of the client to apply it. So styles would be applied just like they used to before the mcss refactor, except now we can have **multiple style files** and they can be **either css or mcss** (oh and the wonderful new dom structure). 

A nice side effect is we can ditch all of that boilerplate `gives: "mcss"` and `readFileSync` repetition in every module. I can see no downside since the mcss wasn't really being applied modularly anyway.

In this branch, the styles are now applied directly in [index.js](https://github.com/ssbc/patchbay/blob/extract-styles-from-depject/index.js#L14). If you require the modules manually, they'll have no styles by default (except for the inline ones).

I know this is a bit of a ...uh.. _sore_ spot right now. But thoughts? 

/cc @ahdinosaur @dominictarr @evbogue @mixmix


mmckegg:
BTW, while this is rather relevant to recent "discussions", this is mostly coming from the point of view of merging the latest `patchbay` into `patchwork-next`. 

I would much rather the styles being seperate than tangled up in the depject logic.

mmckegg:
**One more thing:** once patchbay upgrades to use the latest version of `micro-css` (v1.0.0), all of the classNames will become easily stylable from standard css. 

Also it will also be much easier to use mcss to style things that _weren't written for mcss_ since classes are no longer prefixed.

ahdinosaur:
hmm. personally, i'm most interested in finding a way to do style modules using `depject`, since i like system architectures where patterns are consistently applied, and i feel like pulling styles out of depject (and especially into a new top-level folder) is a new edge case (inconsistency) to how we separate functionality into modules. but this is just my own opinion about software architecture.

>  I can see no downside since the mcss wasn't really being applied modularly anyway.

i agree that the current way we're doing style modules is not serving us, keen to explore alternatives. i'm especially interested in the second proposal in https://github.com/ssbc/patchbay/issues/86#issuecomment-278482264, would appreciate feedback.

this all being said, i'm not against merging this, since we can always iterate on modular styles in the future. :heart:

styles/index.js
```diff
@@ -0,0 +1,17 @@
...
+  if (/\.mcss$/i.test(file)) {
+    result += fs.readFileSync(path.resolve(__dirname, file), 'utf8') + '\n'
+  }
+
+  if (/\.css$/i.test(file)) {
+    additional += fs.readFileSync(path.resolve(__dirname, file), 'utf8') + '\n'
+  }
+})
+
+module.exports = compile(result) + additional
```
mixmix: oooo, this is nice!

mixmix:
I'm a fan of Mikey's second proposal - the style modules which has a bult in reduce which can over-ride.
I think I'm attracted to the idea that if I make some widget and you like it, you can grab it, import it, and have some sensible defaults (like some flex + padding) pulled into your app, and if you want to make it purple you can just extend, or if you want to over-ride you can wreck the lot.

I think where patchbay is currently at might not be the sensible minimum. Would you be interested in pairing on taking the styling I wrote and pulling out the sugary opinion bits and seeing where we could put those. I imagine this might become the first 'theme' and then your preferences would become the second one.

mmckegg:
> I think I'm attracted to the idea that if I make some widget and you like it, you can grab it, import it, and have some sensible defaults (like some flex + padding) pulled into your app, and if you want to make it purple you can just extend, or if you want to over-ride you can wreck the lot.

Yeah, I think this is the right approach. 

I've spent the evening working on porting `patchwork-next` to the latest `patchbay` and I've found that most of the styles are okay (and can just be added to), but that I want to **completely override** the `Message` style. Turns out that this is quite difficult to do with my proposal and **a lot easier with the way it is currently**. That's why it's good to try these things! 😆

mmckegg:
@mixmix and yes definitely interested in collaborating on extracting out the sugary bits in the styling. I've also come across a few DOM issues when I wanted to style the messages. We can talk about those too.

mixmix:
closed - thiis has been superseded

PR #98, opened 2017-04-06
Fix typo in npm start

AljoschaMeyer:


arj03:
I think it should be electro (https://www.npmjs.com/package/electro)

AljoschaMeyer:
Hm, that explains a lot. I'm feeling somewhat stupid right now.

PR #100, opened 2017-04-12
Bring lite back from the dead.

ezdiy:
RFC, this is ugly.

Depends on https://github.com/ssbc/ssb-client/pull/21

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "scripts": {
     "lint": "standard",
     "rebuild": "npm rebuild --runtime=electron --target=$(electron -v) --abi=$(electron --abi) --disturl=https://atom.io/download/atom-shell",
+    "lite": "mkdir -p build && browserify index.js | indexhtmlify --title patchbay > build/index.html",
     "start": "electro index.js"
   },
   "repository": {
@@ -46,9 +47,23 @@
     "ssb-mentions": "^0.1.1",
     "suggest-box": "^2.2.3",
     "text-node-searcher": "^1.1.1",
-    "xtend": "^4.0.1"
+    "xtend": "^4.0.1",
+    "brfs": "1.4.3",
+    "es2040": "^1.2.4",
+    "bulkify": "*",
+    "read-directory": "*",
   },
   "devDependencies": {
-    "standard": "^8.6.0"
+    "standard": "^8.6.0",
+    "browselectrify": "^1.0.1",
+    "indexhtmlify": "^1.3.1"
+  },
+  "browserify": {
+    "transform": [
+      "brfs",
+      "bulkify",
+      "read-directory/transform",
+      "es2040"
+    ]
   }
 }


dominictarr:
oh it's just some deps? I only see changes to package.json. you made it sound like something worse than that. is there stuff not checked in?

*one thing:* you should _never_ use `*` ranges, that is just inviting that module to break on you. you generally want `^` ranges.
If you do `npm install <module> -S` it will save the dep with a sensible range in dependencies field, or if it's a dev dep, use `-D`

mixmix:
hey this is pretty exciting. I'm keen for us to have clear priorities for each client. I'd like to make patchbay that is focused on being a standalone client, but which can also be exported so that it can be used in projects like this.

We're not far off. I think we could make electron an optional dep, and if you check out what's going on in index.js, I think we'd export what was bulk-required ...

patchbay-lite would then import and basically wrap it in this `indexhtmlify` script.

Are you keen?

mixmix:
@ahdinosaur might is looking into this now !

ezdiy:
@dominictarr I know, I use '*' for WIP, -s when final

One problem with brfs is that resulting index.html is 2MB (not minified). Any ideas how to make it not do that and load only restricted subset of things - there are some fancy glob masks for bulk requests, can we pass *that* to brfs somehow?

ahdinosaur:
@ezdiy hey, so @mixmix, @mmckegg, and i are thinking the best approach for the lite client rebirth is to create a separate package that uses the depject modules from `patchcore`, `patchbay`, in a way similar to [the current `patchbay/index.js`](https://github.com/ssbc/patchbay/blob/1e24c51528660ca6d322ca62c6566f9b047c6412/index.js), **plus any special modules to override default behavior to handle cases like https://github.com/ssbc/patchbay/pull/102**.

i'm here to help facilitate that to happen: https://github.com/ssbc/patchbay/pull/102, https://github.com/ssbc/patchcore/pull/11, https://github.com/ssbc/patchcore/pull/12.

ezdiy:
@ahdinosaur Fine by me. I'm no js artisan, I have no idea what depject does (or much of node for that matter). Too much spooky action at a distance for me. Basically, the only thing remaining to be done is to supply config *somehow* - preferably without mucking with node_modules/ as it is done now. As in the previous thread, this is pretty much enough:

```
@@ -6,7 +6,7 @@ exports.create = (api) => {
   var config
   return nest('config.sync.load', () => {
     if (!config) {
-      config = Config(process.env.ssb_appname)
+      config = Config(process.env.ssb_appname, typeof localStorage !== 'undefined' && localStorage.config)
       config.manifest = require('./manifest.json')
     }
     return config

@@ -53,7 +55,7 @@ module.exports = function (name, defaults, argv, parse) {
   ]))
 }
```
Now it depends if you want the typeof atrocity in there, or special-case it to for different npm build target (again, i have no idea how).

ahdinosaur:
@ezdiy aight, how bout you keep finding the hacks necessary to get the lite client working, i'll try and package it up in a way that works with depject. :smile:

ahdinosaur:
made a repo, got the depject bits connected: [`patchlite`](http://github.com/ahdinosaur/patchlite). :smile_cat:

mixmix:
patchlite is merged. Waiting on permissions to publish to npm

dominictarr:
awesome stuff everyone!

PR #104, opened 2017-04-14
Saner start defaults

arj03:
I'm not sure if this is valuable for anyone else. But thought I would try and see, since I got a merge conflict updating my branch :-)


mixmix:
hey I'm not so into the zoom here. Can you unpack what it is you want to see? 
Could be that if there's config you want to persist that this could be a good motivation for building a depject module called `patch-preferences` or something?

arj03:
Sure. I think the title should be default. The other ones would probably be better in some kind of preferences that saves and auto-loads on start.

arj03:
Closing this. Will do a new PR

arj03:
@mixmix See https://github.com/ssbc/patchbay/pull/132 :)

PR #107, opened 2017-04-19
Patch context

dominictarr:
adds a context menu to patchbay, but especially, adds a convienient way to get a viewer.scuttlebot.io link.
it's just a very small addition so it could also be added to patchwork, but i developed it here because I was having trouble running patchwork with my sbot setup.

### DIFF
--- a/main.js
+++ b/main.js
@@ -13,7 +13,7 @@ require('setimmediate')
 // from more specialized to more general
 const sockets = combine(
   // git,
-  {'patch-context': require('patch-context') },
+  require('patch-context'),
   horcrux,
   patchbay,
   patchcore
--- a/main.js
+++ b/main.js
@@ -13,6 +13,7 @@ require('setimmediate')
 // from more specialized to more general
 const sockets = combine(
   // git,
+  {'patch-context': require('patch-context') },
   horcrux,
   patchbay,
   patchcore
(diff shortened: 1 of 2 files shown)


main.js
```diff
@@ -13,6 +13,7 @@ require('setimmediate')
 // from more specialized to more general
 const sockets = combine(
   // git,
+  {'patch-context': require('patch-context') },
```
ahdinosaur: why the wrapping here?

dominictarr: ah, or should it go in patch-context itself?

dominictarr:
I pushed this last night just before I went to bed, it works but I'd like to tidy it up. I think it would be much better if it created a depject socket, message.sync.context or something so that we could easily add other message context actions.

dominictarr:
@ahdinosaur oh yeah, am working on this to support zach's work on the FAQ

mixmix:
Nice work @dominictarr . merged and releases as 7.6.0

Sorry I hadn't seen you'd don't this. I don't have any regular checking of pull requests going as a pattern. D:
I'd actually watched you code the patchwork one, and thought "huh ... I wish he put that into patchbay too"

dominictarr:
@mixmix haha, oh I did this one first because it was easier to access the context menu because patchbay doesn't use `electron-spellchecker`

PR #108, opened 2017-04-21
Fix the size of emojis

arj03:
The emojis are quite big all of the sudden when running with npm start. This makes them the same size as the text. Not sure if the other things was a feature or not, but I like them smaller :)

### DIFF
--- a/app/styles/css/global.js
+++ b/app/styles/css/global.js
@@ -15,6 +15,11 @@ body {
   color: #222;
 }
 
+.emoji {
+  width: 16px;
+  height: 16px;
+}
+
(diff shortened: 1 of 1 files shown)


mixmix:
Merge this ! 
I shifted the size to use `1em`, so that H1's can still have MASSIVE EMOJI ! 

I was kinda quitely enjoying the ridiculousness of the size of the emoji for a bit there, but they needed to be fixed. 
thanks. This is pushed to master, and will go out to npm with the next patch @arj03

arj03:
Thanks @mixmix 👍

PR #114, opened 2017-05-17
handle privacy in blobs

dominictarr:
blobs are currently not encrypted, but since the button is still there, people (reasonably) assume that it _is_ private. At the very least, we should just not show the file attachment on private messages.

I figure that should actually go in the file plugin though. I was disapointed to see that patchbay and patchwork have nearly exactly the same code in messages/html/compose but it's copied into both programs instead of reused.

Shall we shift that into patchcore? @ahdinosaur @mmckegg ?

### DIFF
--- a/message/html/compose.js
+++ b/message/html/compose.js
@@ -20,7 +20,13 @@ exports.needs = nest({
 exports.create = function (api) {
   return nest({ 'message.html.compose': compose })
 
-  function compose ({ shrink = true, meta, prepublish, placeholder = 'Write a message' }, cb) {
+  function compose (opts = {}, cb) {
+    var {
+      shrink = true,
+      meta,
+      prepublish,
+      placeholder = 'Write a message'
+    } = opts
     var files = []
     var filesById = {}
     var channelInputFocused = Value(false)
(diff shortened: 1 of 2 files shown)


message/html/compose.js
```diff
@@ -77,9 +83,10 @@ exports.create = function (api) {
       textArea.value = textArea.value.slice(0, pos) + insertLink + textArea.value.slice(pos)
 
       console.log('added:', file)
-    })
+    }, opts)
```
mixmix: So is the proposal that this opts passed down the info that this is private and then the input template in core decides how to deal with that?

mixmix: Haven't looked but is that an existing core feature or are you suggesting it should exist.
Sounds like a great idea

mixmix:
@dominictarr re disappointment at un-dry code. This is expected, we have little what the best patterns are here, so we're making it up. Some of the missing part to a healthy pattern is a period sync-up and comparing of notes and apps to adjust courses and refactor.
I'm all for tidy code - that's why I've spent almost all of my energy trying to refactor things like patchbay and git-ssb to make them easier to maintain and extend. I notice that for the moment I'm feeling burnt up on tidying, and want to build some cool shit. I expect an oscillation between cool shit and tidying might be healthy for me and for the projects.

PR #115, opened 2017-05-30
fix font paths using resolved absolute paths

ahdinosaur:


mixmix:
such a tidy little fix <3

Issue #116, opened 2017-05-31
fs.readFileSync() in client code


regular:
hi, I tried to build index.html to serve it to a browser using

```
browserify main.js | indexhtmlify --title patchbay > build/index.html
```
(as mentioned in the Readme).

The resulting file however contains  [this call to fs.readFileSync](https://github.com/ssbc/patchbay/blame/53d03c145268665aa93e60777b4153a8a0c4dca5/app/styles/css/font-awesome.js#L8) and the browser actually hits that code.

I might be totally wrong about how this is meant to work :/

dominictarr:
hmm, brfs is ment to be able to handle that but I think @ahdinosaur broke it in this commit. https://github.com/ssbc/patchbay/commit/53d03c145268665aa93e60777b4153a8a0c4dca5

ahdinosaur:
hey @regular, if you want to run `patchbay` in the browser, i recommend [`patchlite`](https://github.com/ssbc/patchlite) which `patchbay` + a few modules to make everything work in the browser.

@dominictarr `patchlite` hasn't worked with the latest `patchbay` since those fonts were added, we've kept `patchlite` at the latest working commit: https://github.com/ssbc/patchlite/pull/2. i'm hoping to make it work with the latest `patchbay` again and that change is one step closer.

dominictarr:
@ahdinosaur what does styleResolve do?

ahdinosaur:
@dominictarr https://github.com/stackcss/style-resolve#usage

did a pass at making this work: https://github.com/ssbc/patchbay/pull/117

mixmix:
Cool, can look at art hack tmrw.

christianbundy:
I see that #117 was merged, does that mean that this is ready to close?

PR #117, opened 2017-06-06
use `require-style` to require font styles in a patchlite friendly way

ahdinosaur:
haven't actually tested this on `patchlite` yet, but might work :smiley:

https://github.com/ssbc/patchbay/issues/116


ahdinosaur:
this works! @mixmix 

although `patchlite` is still unable to upgrade to the latest, since now it's stuck on an issue with using the new flume plugins from the browser. but at least one less blocker. :smile_cat:

PR #118, opened 2017-06-13
WIP - patch-router

mixmix:
This is a gesture at how we might perhaps do depject-style routing for the patch-core eco-system.

## Major changes

- there is a common `router.sync.routes` that any module can depect into
  - each patch-module contributes its chunk of topic-related routes
- views now live in their associated topics e.g. `channel.html.page`
  - no more weird cramming pages in under `app.html.page`
  - might mean a lot more "topics" e.g. public, private, notifications, profile (== feed?)

@mmckegg @ahdinosaur @dominictarr if this was to be common perhaps we would just have the  `router.sync.router` module in patch-core.
 
---

**Not coded in this WIP**

- all the page changes (I've only shown moving the `blob` page as an examples)
- catching of clicks and feeding into the router
  - note sheet-router has its own click catching


ahdinosaur:
hey @mixmix, nice pull request to keep this conversation moving. :tractor: 

what's the reasoning behind doing it this way rather than what i proposed in %mgICtBOBXbc7tJth60u4pA1Jg7qN/BtL4Fe8ckhT8mk=.sha256 and %JpQ9iRrd+MT5L35MBandTpFHJ0w6inxQoAbUmB0BV0I=.sha256?

> there is a common `router.sync.routes` that any module can depect into

seems the pro is there is one module that handles _all_ routing via a reduce, so it's easy to remove undesired routes or add a bunch. the con is that it's all one module, so all your routing code is necessarily entangled together.

> - views now live in their associated topics e.g. `channel.html.page`

if the routes modules will specify these modules by name, what's the benefit of having them in a special module path forcing each topic to have only one page? i'm not sure this is a good pattern, too many topics can be messy, many topics will have more than one page (for example, git). could be either a new type, but really i feel that it makes more sense for pages to be html modules that are referenced by routes.

mixmix:
@ahdinosaur I misunderstood the API suggestion I think. I didn't see you were meaning to make that API all accessible via depject. I think that part is a good idea and should be accessible via sheet-router. (haven't looked in detail)

Note there is  `router.sync.router` which collects all routes and builds the actual router. `router.sync.routes`  is different and is the 'patch' (?) level routing. I imagined there being one of these for patch-bay, one for patch-hub, one for patch-gatherings. I think the reason this is important is that ORDER matters for things like sheet-router, so having the order controlled on a per-patch is a good thing. I personally also like being able to look at routing once for each patch. Also note that not all the routing is 'tidy' in that all the key-based routes are kinda 'special' and have to be lumped together (using sheet-router at least)

I'm not sure what the best pattern is for the topic pages. I did notice that having tasted an alternative that I really don't like the pattern of `app.html.page` hiding a bunch of stuff. I don't think the pattern needs to be `git.html.page` , I'd imagined it would become things like `git.html.index` and `git.html.repo` - actually with routing the naming doesn't matter that much, the router can point at any html view.

mixmix:
I mainly want a consistent and predictable pattern - like should pages be all in one place or to be expected to be in a number of places at the same depth.

I like the idea of `blob.html.show` as a view.
oooo... what about : 

```js
    const routes = [
      ['/public',        () => api.page.html.public()],
      ['/private',       () => api.page.html.private()],
      ['/notifications', () => api.page.html.notifications()],
      ['/profile',       () => api.page.html.profile({ id: myId })],
      ['/:key', (params) => {
        const { key } = params
        if (isBlob(key))    return api.page.html.blob(params)
        if (isFeed(key))    return api.page.html.profile(params)
        if (isMsg(key))     return api.page.html.thread(params)
        if (isChannel(key)) return api.page.html.channel(params)
      }]

```

ahdinosaur:
@mixmix 

so the current proposal is for one plug per 'patch': `router.sync.routes`, which is a reduce that receives the `sheet-router`-compatible routes tree and returns a new routes tree.

an alternative proposal is `router.sync.route`, which is a first that returns a path and a handler. (i think handling "the key-based routes are kinda 'special'" is not a problem, given a path and a handler we can add the handler at the path in the routes tree.)

i think you're right that the former is better, because

- order matters (for example: if a 'patch' exports two plugs: '/cats' and '/cats/:id')
- each plug can update the routes as necessary (same reason we moved from first to reduce with the css plugs)
- routes per 'patch' are in a single place, like a site map

so :+1: 

> I don't think the pattern needs to be `git.html.page` , I'd imagined it would become things like `git.html.index` and `git.html.repo` - actually with routing the naming doesn't matter that much, the router can point at any html view.

yes! this is what i was trying to say. :smile_cat: 

> I mainly want a consistent and predictable pattern - like should pages be all in one place or to be expected to be in a number of places at the same depth.

how do you feel about the `git` 'patch' exporting `router.sync.routes`, which references `git.html.index` and `git.html.repo`? otherwise would recommend adding `page` as a "type", i really don't think it's a "topic". this is how i would do it in `dogstack` land (we already have `components` and `containers`, am actually thinking about `pages`).

mixmix:
cool, feels like this is coming together.
So do you mean to propose something like `app.page.profile` ? 

I was definitely looking at the things in `app.html.*` and thinking "these are all fucking partials ... or helpers ..  i want them gone"  (apart from the root element `app.js`)

Is html to vague? should it be `app.view.profile` and `app.partial.tabs` ... mmm could be an improvement... it feels a little leaky compared to the cleanliness of `obs pull sync async`

ahdinosaur:
> Is html to vague? should it be `app.view.profile` and `app.partial.tabs` ... mmm could be an improvement... it feels a little leaky compared to the cleanliness of `obs pull sync async`

`html` is vague, which is a strength _and_ a weakness. `view` and `partial` make sense, it's only a problem if there are html modules which are neither a view nor a partial, and if it's very clear whether something is a view or a partial. what is a view and what is a partial?

mixmix:
we get to define that I guess. I like `view` over `page` as a collected largest chuck I'm going to interact with. But I don't really care. Good point about some of the other misc ones .. like the scroller or the tabs .. wtf are they ...they're html which has an API ....

mixmix:
@ahdinosaur here's where I'm up to. 
This is mostly working with https://github.com/ssbc/patchcore/pull/24

I've thought about your proposed API for history pushing etc. I think I need to consider more how the history would work with multiple tabs. I also think this is useable without that straight away.

mixmix:
oh - `sheet-router` does not like message keys ... that's annoying. hacking....

ahdinosaur:
> message keys 

what are "message keys"?

mixmix:
like `%ZhbqNxg+gdznRGVSET1db5PbiMGjqfzoGI6qf6Vgoz0=.sha256`  I've got a hack which adds a / as a prefix, but then url encoding / decoding kicks in. 
I'm digging in sheet-router but can't see where that's happening. Suspect it might be upstream in wayfarer.

Off to bed though now

ahdinosaur:
@mixmix we can't use raw message keys in the routes, since `%` (the sigil) is a special character to means a url encoding and `/` (which is part of base64) often means a delimiter between routes. we might want to replace the sigil with a route path (`/message/:keyWithoutSigil`) and use something like [`base64-url`](https://www.npmjs.com/package/base64-url) to escape the message key.

mixmix:
hey @mmckegg I've refactored this (and patchcore branch) to run object-based routing. As in locations are represented by objects.

_Reminder that we need something more complex for routing in patchcore to build patch-hub links nicely...._

In patchcore there's a normaliser which catches e.g. feedId strings and maps them to form `{ feed: id }`.

I've followed a pattern similar to sheet-router - the routes is an array of route entries. The first entry is a function which validates whether a location passed to it is a match. This feels super flexible. 
The second is a function which will be passed the location object if it's a match.

Whadya reckon?

dominictarr:
I notice that there is now one file that has a list of all routes, instead of having each page hook a standard api?
This makes it harder to "just merge" in a new feature, like, it's a one liner to add patch-gatherings... or is there something I am missing?

like, %JpQ9iRrd+MT5L35MBandTpFHJ0w6inxQoAbUmB0BV0I=.sha256

mixmix:
Each patch-app might have a single (or multiple routes gives). I think
subtle is better in a patch because you can specify order and read about
routes in one place.

Each patch app reduces the routes, so adding patches is easy

app/html/app.js
```diff
@@ -45,21 +38,24 @@ exports.create = function (api) {
...
-    var { container: errorPage, content: errorList } = api.app.html.page('/errors')
-    window.addEventListener('error', ev => {
-      if (!tabs.has('/errors')) tabs.add(errorPage, true)
+    // var { container: errorPage, content: errorList } = api.router.sync.router('/errors')
+    // window.addEventListener('error', ev => {
+    //   if (!tabs.has('/errors')) tabs.add(errorPage, true)
 
-      const error = api.app.html.error(ev.error || ev)
-      errorList.appendChild(error)
-    })
```
mixmix: revisit this

mixmix:
This is merged. @ahdinosaur felt a need to keep this moving. Also, I've put this pattern through it's paces for that recent Chinese contract, and found it feels really great.

I'm sure there's lots to learn still, but I'm confident that this is a necessary good step in the right direction, and that changes from this point will actually be _easier_ because we've decoupled some concerns.

app/html/app.js
mixmix: :white_check_mark: fixed.

PR #119, opened 2017-06-17
fulltext is not optional as search will not work if is not installed

arj03:


mixmix:
mmm, thought I tested this. thanks @arj03

PR #121, opened 2017-06-29
fix compose overwriting observable meta

mmckegg:
This fixes an issue where a subsequent reply would get stuck replying to the same post as the previous.

Possibly fixes other weird issues.

Corresponding fix in Patchwork: https://github.com/ssbc/patchwork/commit/6a4dd175b0e6195afeec9be9edf7a1e0723dfce0


mixmix:
thanks @mmckegg !

PR #122, opened 2017-06-29
bump ssb-mentions for channel hashtags

mmckegg:
I just noticed that patchbay is not generating _channel mentions_. This means that posts like [this one](https://viewer.scuttlebot.io/%250q9iOAgBixNFjeLO1UZefzCX1Xsv09Pl2C%2Fl2ACjDQQ%3D.sha256) won't show up under the mentioned channels. 

This is because it is using an old version of `ssb-mentions`. PR updates to the latest version.


mixmix:
catch ! 
thanks @mmckegg

PR #128, opened 2017-08-05
Need electron and electro too

alanz:
From installing on a new node setup


mixmix:
thanks for this. I'd written a script `npm run setup`, but maybe forgotten to document. :)

PR #129, opened 2017-08-09
added display of your current funds

dangerousbeans:


arj03:
I really like this, but without a way of added these transactions in patchbay its a somewhat half experience. @dangerousbeans are you planning on doing that as well?

dominictarr:
@arj03 that is a good point, but on the other hand, I think our policy should be merge merge merge! if it doesn't break stuff, merge it. obviously we'll need making transactions for this part to make sense, but we can still add this. that said, if there was a way to make a transaction in the PR it would be an easier argument!

mixmix:
I think there's a minor conflict with the ssb-chess version. I can fix that and merge this when I get up

mixmix:
Hey I've pull this, merged master, and smoke-testing it I'm getting a bug.
I've added you as a collaborator on this repo @dangerousbeans and opened a new local PR : https://github.com/ssbc/patchbay/pull/131

PR #132, opened 2017-08-31
Save window settings

arj03:
It now uses patch-settings :)

### DIFF
--- a/app/html/app.js
+++ b/app/html/app.js
@@ -59,22 +59,27 @@ exports.create = function (api) {
       addError(ev.error || ev)
     })
 
-    window.addEventListener('resize', function(e){
-	var wc = electron.remote.getCurrentWebContents()
-	wc && wc.getZoomFactor((zf) => {
-	    api.settings.sync.set({
-		'electron-zoomFactor': zf,
-		'electron-windowBounds': electron.remote.getCurrentWindow().getBounds()
-	    })
-	})
+    // TODO - extract this to keep patch-lite isolated from electron
+    window.addEventListener('resize', () => {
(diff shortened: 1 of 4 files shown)


arj03:
Right, namespacing is a good idea.

app/html/app.js
```diff
@@ -56,6 +59,29 @@ exports.create = function (api) {
       addError(ev.error || ev)
     })
 
+    ////// TODO - extract this to keep patch-lite isolated from electron
+    const { getCurrentWebContents, getCurrentWindow } = electron.remote
+    window.addEventListener('resize', () => {
+      var wc = getCurrentWebContents()
+      wc && wc.getZoomFactor((zf) => {
```
mixmix: @arj03 I've set it up so the settings takes objects - I've changed it to use this for name-spacing

app/html/app.js
```diff
@@ -56,6 +59,29 @@ exports.create = function (api) {
...
+          }
+        })
+      })
+    })
+
+    var zoomFactor = api.settings.sync.get('electron.zoomFactor')
+    if (zoomFactor)
+      getCurrentWebContents().setZoomFactor(zoomFactor)
+
+    var bounds = api.settings.sync.get('electron.windowBounds')
```
mixmix: here's what getting the settings out looks like (I think I mentioned in docs that it follows lodash/get style)

mixmix:
works well, I can see why you're keen for this !

mixmix:
Merged and published. Nice work @arj03 :fireworks: :pineapple:

PR #133, opened 2017-09-01
Embedded sbot

mixmix:
this embeds an sbot in patchbay, so you don't have to run an alternate scuttlebot fork for private messages, and don't need to worry about keeping that stuff up to date.

I'm definitely interested in this being an optional setup - as in I want to support people who : 
- like to run their own standalone sbot + electro
- want to require patchbay (e.g. patchlite)

I think some of the deps should be moved to devDependencies ... haven't played with that much yet (had a hard time getting this going because `electron .` was reading `main.js` instead of index.js D:)

## Still needs 

- README update

cc @ahdinosaur


mixmix:
I'm merging this.
Moving towards building installers for patchbay to see what that's like

Happy0:
Nice! :D

PR #135, opened 2017-09-05
On select from dropdown go directly to result, don't wait for enter

arj03:
Thought this would be more intuitive


mixmix:
great suggestion - one of those suggestions that you wonder how you lived without.

I just merged a change which tracks which tabs you've been using, so when you close them you go back to the ones you were last in ... similarly simple idea, but hoping it'll also make a difference

PR #138, opened 2017-09-08
save filters

arj03:
Another candidate for saving. I use Only people I follow a lot and don't have to turn it on every time I start the app :)

### DIFF
--- a/app/html/filter.js
+++ b/app/html/filter.js
@@ -4,6 +4,8 @@ const Abort = require('pull-abortable')
 const pull = require('pull-stream')
 const addSuggest = require('suggest-box')
 const { isFeed } = require('ssb-ref')
+const some = require('lodash/some')
+const get = require('lodash/get')
 
 exports.gives = nest('app.html.filter')
 
@@ -11,7 +13,7 @@ exports.needs = nest({
   'about.async.suggest': 'first',
   'contact.obs.following': 'first',
   'keys.sync.id': 'first',
-  'settings.sync.get': 'first',
+  'settings.obs.get': 'first',
   'settings.sync.set': 'first'
 })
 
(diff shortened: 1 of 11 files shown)


mixmix:
great feedback. I'll have a look at what you've done when I can. I was thinking about this.
my thoughts so far are that you might want to save settings for : 
- each type of page
- use a white-list type approach (given the proliferation of different message types?).

arj03:
I actually started out wanting to do channel filters, but started with something more simple. Also it would probably be a good idea to mark the filter icon as filtering, so that it's easier to see that you are filtering. In case you forgot (more relevant with save :))

arj03:
Right so this now shows when filters are active or not. In TOXIC green as I can't pick colors for shit :)

I actually don't use the author thing much, why not just use the top right search? So I'm thinking of just removing that.

Yeah the type is interesting. I think I'll add a chess filter :)

app/html/filter.js
```diff
@@ -23,15 +25,20 @@ exports.create = function (api) {
...
-    const showContact = Value(false)
-    const showChannel = Value(false)
-    const showPub = Value(false)
+    const showPost = Value(api.settings.sync.get('filter.showPost') || true)
+    const showAbout = Value(api.settings.sync.get('filter.showAbout') || true)
+    const showVote = Value(api.settings.sync.get('filter.showVote') || false)
+    const showContact = Value(api.settings.sync.get('filter.showContact') || false)
+    const showChannel = Value(api.settings.sync.get('filter.showChannel') || false)
+    const showPub = Value(api.settings.sync.get('filter.showPub') || false)
+    const showChess = Value(api.settings.sync.get('filter.showChess') || true)
```
mixmix: couple of things here : 
- get has a built in fallback. Go read lodash/get. Basically this could be : api.settings.sync.get('filter.showChess', true)
- I think I want to set a default for the app settings .. might put that in `app.sync.settingsDefaults`

mixmix: - `api.settings.obs.get` actually get your an observeable value

arj03:
Thanks for the refactor. Its a lot more DRY now. I fixed some minor things in your commit :)

mixmix:
Awesome! Look forward to checking it out

arj03:
Got channel filters working. For me this is a soft block. Like at one time when a lot of Russians came in, they were nice enough to keep to #russian, so having the ability to exclude that channel would be very welcome for me :)

I'm pretty happy with this PR now, except for the color of the filter when its active :smile:

mixmix:
Hey arj, haven't forgotten, want to twist this, and also feeling pulled a
few ways.
I'll aim to review and get this merged this week.

Keep up the great work producing functional improvements. I might
prioritise your patchcore pr over this because I think lots of people could
benefit (unless it's already merged!)

arj03:
Thanks @mixmix, always happy with your improvements :)

mixmix:
I rebased this, just to see if that workflow is good.

I don't think the styling is perfect. I fell into a rabbit hole with the average page layout. I decided to not refactor the pages layout (as you can see I started to with the `channel` page) and instead have something minimal which just says 'custom' in front of the filter icon when there are custom filters active ... 

Let's get into master and enjoy it. If it's confusing we'll come back to it.
Nice work @arj03

PR #139, opened 2017-09-08
Channel subscription

arj03:
This is a WIP pull request. I was missing this functionality from patchwork so thought I would port it over. I'm not very good with styling so I could use some help if anyone wants to. 

- I have added a header to the channel page. I needs to be formatted as such, right now it has no styling. I'm also not sure if we should write is subscribed there, or if it should be an unsubscribe button, if you are already subscribed.
- When subscribing to a channel the button doesn't change, before merging some kind of visual feedback needs to happen. I'm not sure what the best way is.


arj03:
Related https://github.com/ssbc/patchcore/pull/28

arj03:
Last commit should take care of the points mentioned above. So this is ready for inclusion if anyone else finds it useful :hamster:

arj03:
@mixmix I merged in the related patchcore commit. Do you have an opinion about this PR?

mixmix:
@arj03 thanks for adding this. I've brought this up to date.
I did a couple of things to the code you wrote: 
- re-jigged the styles for the channel page. This probably needs an overhaul and design, so we can get a btter layout and DOM strucutre in the future. In the meantime there's a pattern I've enjoyed in another project that I've added `Page -channel`
- I used `mutant` magic to prune your code down. I think you should read when I wrote to learn more about mutant, it's really nice once you get it

app/page/channel.js
```diff
@@ -19,9 +23,25 @@ exports.create = function (api) {
...
+        channel: channelName,
+        subscribed: !subscribed()
+      })
+    }
+
+    const subscribeButton = h('Button -subscribe',
+      { 'ev-click': toggleSubscription },
+      when(subscribed, 'Unsubscribe from channel', 'Subscribe to channel')
+    )
+
```
mixmix: Check out `toggleSubscription` and `subscribeButton` to see what I've changed.
Ask questions if there's anything that seems too magical here or unclear :)

p.s. I learnt more about mutant from your code here - it didn't know obs.subscribed.has was an obs as well !

mixmix: @arj03  ^^

mixmix:
Nice Pull request, this was much needed !

app/page/channel.js
arj03: Yes makes sense. Still quite new to mutant, but this is clearly DRY :)

PR #140, opened 2017-09-10
Inbox

mixmix:
An inbox for patchbay (and in the future, the patchcore ecosystem)

selection_377

works with keyboard shortcuts and click !


arj03:
I was trying to test this. You tweaked the overall css as well to fill up the screen?

I guess this depends on the inbox branch of scuttlebot as well?

mixmix:
No this is using the private branch of scuttlebot. See you've tried that
too. Not sure what that error you got was, I think I've seen that before..

I tweaked a couple of minor css things and upgraded mcss

PR #142, opened 2017-09-11
better mention search

mixmix:
use some simple regex matching.
The has the effect of checking if the word typed occurs anywhere _within_ strings, as opposed to having to be right from the start.

e.g. `dino` will match ah**dino**saur
e.g. `cat` will match crying_**cat**


mixmix:
turns out this was a bad idea, because if you type `:)` that makes BBAAAAD regex.

swapped this out for `indexOf` which is much simpler : dfae69d4d300396177503df2b63ac8c17c7ffb7d

PR #149, opened 2017-09-27
Bump electron to 1.7 as it is now stable

arj03:


mixmix:
thanks for noticing this - I've skipped over 1.7 and straight to 1.8.1 (this is was patchwork has been using and seems to be stable). I think there were some unpleasent mystery hanging issues for 1.7 that patchwork was hitting real bad, so keen to avoid those :P

PR #150, opened 2017-10-19
Fix invisible confirm button for changing names

arj03:
Seems to be white on white. Not sure how it was broken, but this seems to fix it.


mixmix:
Great catch. thanks !

PR #152, opened 2017-10-22
display who a profile is blocking, as well as who they are blocked by

dangerousbeans:
screen shot 2017-10-22 at 2 17 24 pm

### DIFF
--- a/contact/html/relationships.js
+++ b/contact/html/relationships.js
@@ -13,6 +13,7 @@ exports.needs = nest({
   'contact.obs.followers': 'first',
   'contact.obs.following': 'first',
   'contact.obs.blockers': 'first',
+  'contact.obs.blocking': 'first',
   'keys.sync.id': 'first'
 })
 
@@ -56,6 +57,7 @@ exports.create = function (api) {
 
     const { unfollow, follow, block, unblock } = api.contact.async
     const blockers = api.contact.obs.blockers(id)
+    const blocking = api.contact.obs.blocking(id)
     const ImBlockingThem = computed(blockers, blockers => blockers.includes(myId))
 
     return h('Relationships', [
@@ -89,6 +91,14 @@ exports.create = function (api) {
           ])
         ])
       ),
+      computed(blocking, blocking => {
+        if (blocking.length === 0) return ''
+
+        return h('div.blocking', [
+          h('header', 'Blocking'),
+          h('section', blocking.map(imageLink))
+        ])
+      }),
       computed(blockers, blockers => {
         if (blockers.length === 0) return ''
 


arj03:
@dangerousbeans I have tested this an it works great. Cool stuff! Not sure if I want it at the top above Friends. Under Followers would be better I think.

dangerousbeans:
@arj03 I agree, though this is just adding "Blocking" beside "Blocked By" - could move both down?

evbogue:
This is cool, and I should implement this in minbase. But it made me wonder: who is blocking me, and what did I do?

evbogue:
But I'm in good company, because if anyone is blocking @mixmix's dad might as well block me too.

arj03:
@dangerousbeans I think they are asymmetric. If someone is blocked I wanted to see it next to their name and picture. But if someone is blocking someone else, I don't need to see it right away. It is less important than who their friends are. I think ;)

arj03:
Yeah @evbogue. Currently the ban is the big hammer. We need something more fine grained. Would be really nice to be able to tag posts as offence as that it probably the first step towards banning ;) This has been talked about, I think we just need someone to actually implement it :)

evbogue:
@arj03 I think flag and downvote have both been discussed. The original Patchwork had flag for a brief moment in time, if I remember correctly.

I find myself wondering if people want their blocklists to be public on their profiles? I'm not the person in the screenshot above, but I imagine that the person in question was trying to distance themselves from me when they went from following me to blocking me. Having my profile photo on their profile page would probably lead to more questions than answers.

evbogue:
sorry, my trackpad freaked out and hit the close button.

arj03:
The info is public anyway. The reason why I think this is so great is that maybe you blocked somebody at one point but forgot about it and with this its really easy to see who you blocked.

dangerousbeans:
right now the blocks are broadcast very publicly but the 'blockers' are hidden in all the UI

if you were relying on a social connection to connect you to another group, that link may be lost if someone decides to block people

Or, if a pub decides to block big ranges of people than maybe you'd want to be able to see that

Anyway I think we should merge this but also redesign the UI of the profile page

arj03:
Sure I think this is good as well, are you okay with me moving it down a bit?

dangerousbeans:
@arj03 yis

mixmix:
Cool!

I started a branch somewhere to make the different relationships like tabs
you can click through. This way we can give a lot of detail without having
to jam up the whole interface.

I'm keen to get "friends in common" in here too

dangerousbeans:
hellz yes for friends in common

PR #154, opened 2017-11-06
Use new version of ssb-chess which does not require sqlite3

Happy0:
This version of ssb-chess does not require sqlite3. Instead, it uses `flumeview-reduce` in the `ssb-chess-db` scuttlebot plugin.

Before this is merged I'd like someone to test / try out the following:
* Send me a game invite and tell me you've done so I can make sure 'received' invitations are still displayed.
* Make sure the 'observe' tab still looks sensible.
* Make sure the 'my moves' and 'games' tabs still look as you expect.
* View a player's profile and make sure the list of finished games look sensible.

### DIFF
--- a/background-process.js
+++ b/background-process.js
@@ -20,6 +20,7 @@ var createSbot = require('scuttlebot')
   .use(require('ssb-fulltext'))
   // .use(require('ssb-ebt'))
   .use(require('ssb-ws'))
+  .use(require('ssb-chess-db'));
 
 // pull config options out of depject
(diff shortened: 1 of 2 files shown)


Happy0:
@mixmix I hope you're okay with me requiring my new `ssb-chess-db` plugin? The code is here: https://github.com/Happy0/ssb-chess-db/blob/master/index.js

Happy0:
Just going to merge this to make it easier for people to test (so they don't have to check out my remote, etc.)

PR #161, opened 2017-11-19
add automatic draft saving to composer

cryptix:
nearly identical to ssbc/patchwork#667

### DIFF
--- a/message/html/compose.js
+++ b/message/html/compose.js
@@ -19,7 +19,7 @@ exports.needs = nest({
 exports.create = function (api) {
   return nest({ 'message.html.compose': compose })
 
-  function compose ({ shrink = true, meta, prepublish, placeholder = 'Write a message' }, cb) {
+  function compose ({ location, shrink = true, meta, prepublish, placeholder = 'Write a message' }, cb) {
     var files = []
     var filesById = {}
     var channelInputFocused = Value(false)
@@ -56,7 +56,11 @@ exports.create = function (api) {
     })
 
     var draftPerstTimeout = null
-    var draftLocation = resolve(meta).root || '/public'
+    var draftLocation = resolve(meta).root
+    if (!draftLocation) {
+      draftLocation = location.page
+      if (draftLocation === '/channel') draftLocation = draftLocation + ':' + meta.channel
+    }
     var textArea = h('textarea', {
       'ev-input': () => {
         hasContent.set(!!textArea.value)
--- a/message/html/compose.js
+++ b/message/html/compose.js
@@ -56,14 +56,13 @@ exports.create = function (api) {
     })
 
     var draftPerstTimeout = null
+    var draftLocation = resolve(meta).root || '/public'
     var textArea = h('textarea', {
       'ev-input': () => {
         hasContent.set(!!textArea.value)
         clearTimeout(draftPerstTimeout)
         draftPerstTimeout = setTimeout(() => {
-          // TODO: /public might not be the only place where there is no root?
-          let where = resolve(meta).root || '/public'
-          api.drafts.sync.set(where, textArea.value)
+          api.drafts.sync.set(draftLocation, textArea.value)
         }, 200)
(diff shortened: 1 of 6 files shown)


message/html/compose.js
```diff
@@ -53,8 +55,17 @@ exports.create = function (api) {
...
 
+    var draftPerstTimeout = null
     var textArea = h('textarea', {
-      'ev-input': () => hasContent.set(!!textArea.value),
+      'ev-input': () => {
+        hasContent.set(!!textArea.value)
+        clearTimeout(draftPerstTimeout)
+        draftPerstTimeout = setTimeout(() => {
+          // TODO: /public might not be the only place where there is no root?
+          let where = resolve(meta).root || '/public'
```
mixmix: There are a couple of cases I think you've identified well : 
- **totally new thread**: 
  1. `/public`
    - I think you're where = `/public` might be fine?
  2. `/private`
    - I think we could/ should make `meta` observeable and check for if there are `recps` loaded in. Making it observeable would mean we wouldn't be leaning on the pre-publish to add `recps` .. then we could use `where = serialize(recps, '/private')` ?  I think this might change when I finally make a nice composer for `/inbox` 
    - I'll see if I can get that working ... 
- **extending an existing thread**
  1. a straight reply - using root works fine `root`
  2. a nested reply or a fork - needs `root` and the new `fork` which I'll be implementing (maybe needs everything by `branch` from meta?

I think serialising or hashing meta would be a good `where`, because the meta describes the type of message, the root, and whether it's a nested reply in some implementations.

Buuut.. the branch might change, so maybe we want where

message/html/compose.js
```diff
@@ -64,6 +75,12 @@ exports.create = function (api) {
     })
     textArea.publish = publish // TODO: fix - clunky api for the keyboard shortcut to target
 
+    // load draft
+    let draft = api.drafts.sync.get(resolve(meta).root || '/public')
```
mixmix: The `where` logic is repeated, could move it up, and call it something like `draftLocation`

message/html/compose.js
```diff
@@ -64,6 +75,12 @@ exports.create = function (api) {
...
+    if (typeof draft === 'string') {
+      textArea.value = draft
+    }
+
```
mixmix: :fire: - We need to clear drafts on publish!

The perfect place is in the `done` callback, which is inside the `publish` function near the bottom

mixmix:
Added you as a contributor to patchbay to make jamming together easier ! 
https://github.com/ssbc/patchbay/invitations

Can't say how excited I am to find this contibution, it's been on my wishlist :heart: :two_hearts:

message/html/compose.js
mixmix: with the `private` page... making meta observealbe looks like a good refactor, but not one I hav etime to do well right now.

Minimal solution ? 

pass the `location` into compose directly : 

```
  function compose ({ location, meta, prepublish, shrink = true, placeholder = 'Write a message' }, cb) { 

    var where = location || meta.root+meta.fork  // or something
```

In patchbay, routing is done with 'objects' - you'll see each page is passed a `location`, and this should be an object like `{ page: '/inbox' }`  or `{ key: '%asd1ny287xkjxw' }` or (in the future) `{ msg: '%asd1ny287xkjxw', fork: '%gga0DD2342wd }`

cryptix:
there you go, @mixmix! I think I even got the `location` passing like you described!

cryptix:
@mmckegg how do you feel about this addition? i'll port it like this to patchwork too then.

mixmix:
hey @cryptix haven't forgotten this, been busy and want to make a couple of tweaks before merging this. 

I duplicated this PR on a local branch so we can all push to it https://github.com/ssbc/patchbay/pull/163
CLOSING this PR in favour of that one.
Going to make a couple of tweaks and mention you to show you what I've changed and if that's all good, then merge

PR #163, opened 2017-11-23
add automatic draft saving to composer (local branch)

mixmix:
This is duplicate of https://github.com/ssbc/patchbay/pull/161 except on a branch in this repo.
I intend to make a couple of tweaks before merging

### DIFF
--- a/app/page/channel.js
+++ b/app/page/channel.js
@@ -39,7 +39,9 @@ exports.create = function (api) {
       when(subscribed, 'Unsubscribe from channel', 'Subscribe to channel')
     )
 
-    const composer = api.message.html.compose({ meta: { type: 'post', channel: channelName } })
+    const composer = api.message.html.compose({
+      location: { page: '/channel' },
+      meta: { type: 'post', channel: channelName } })
     const { filterMenu, filterDownThrough, filterUpThrough, resetFeed } = api.app.html.filter(draw)
     const { container, content } = api.app.html.scroller({ prepend: [subscribeButton, composer, filterMenu] })
 
--- a/app/page/channel.js
+++ b/app/page/channel.js
@@ -40,7 +40,7 @@ exports.create = function (api) {
     )
 
     const composer = api.message.html.compose({
-      location: { page: '/channel' },
+      location,
       meta: { type: 'post', channel: channelName } })
     const { filterMenu, filterDownThrough, filterUpThrough, resetFeed } = api.app.html.filter(draw)
     const { container, content } = api.app.html.scroller({ prepend: [subscribeButton, composer, filterMenu] })
--- a/main.js
+++ b/main.js
@@ -20,7 +20,9 @@ const patchbay = {
     config: require('./config'), // shouldn't be in here ?
     contextMenu: require('patch-context'),
     suggestions: require('patch-suggest'),
-    inbox: require('patch-inbox'),
+    settings: require('patch-settings'),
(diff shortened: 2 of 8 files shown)


app/page/channel.js
```diff
@@ -39,7 +39,9 @@ exports.create = function (api) {
       when(subscribed, 'Unsubscribe from channel', 'Subscribe to channel')
     )
 
-    const composer = api.message.html.compose({ meta: { type: 'post', channel: channelName } })
+    const composer = api.message.html.compose({
+      location: { page: '/channel' },
```
mixmix: note we can and should just use the location on line 22 - it has the correct channel name built in.

main.js
```diff
@@ -36,6 +36,7 @@ const sockets = combine(
   require('patchbay-book'),
   // require('patch-network),
   require('patch-settings'), // might need to be in patchbay
+  require('patch-drafts'),
```
mixmix: [minor] : given how woven in this module is, it needs to be part of the patchbay set of modules up on line 9

message/html/compose.js
```diff
@@ -53,8 +55,20 @@ exports.create = function (api) {
       title: when(meta.channel, 'Reply is in same channel as original message')
     })
 
+    var draftPerstTimeout = null
+    var draftLocation = resolve(meta).root
```
mixmix: I think the draft loction should just be `JSON.stringify(location)`  (or the hash of that). That means it will be unique per page. Could merge in meta as well?

This I need to check more closely to remind myself what's in meta and location on a few of the pages

mixmix: would also simplify the logic below I think

message/html/compose.js
```diff
@@ -178,7 +198,10 @@ exports.create = function (api) {
       function done (err, msg) {
         publishBtn.disabled = false
         if (err) throw err
-        else if (msg) textArea.value = ''
+        else if (msg) {
+          textArea.value = ''
+          api.drafts.sync.set(draftLocation, '')
```
mixmix: Need to check your module. I think we actuallt want to delete the draft. If this is the way you trigger that using your set api, I'd be suprised.

I would expect `api.drafts.sync.set(draftLocation, null)` would actually hard wipe the key in the store.
This is just knit-picking, but will help us from having a million keys with no content in them any more

cryptix: 🔥 this is true! empty string isn't special cased. I did not forsee the acumulation of superseeded drafts just wanted to blank it out and yes, it would get messy in the long run and also not work for a `/drafts` page.

special casing `null` would also be weird though, i think and adding a `.remove(key)` function totally sounds reasonable to me.

message/html/compose.js
cryptix: my thinking was to have keys be something humanly understandable (`/channel#name` would have been better than `:` looking at it now) or already usable by the code (we can already extract titles from message keys).

The only problem I see is having multiple drafts for one channel. that might actually be a requirement but I still find hashes messy here since they don't offer any meaning.

package.json
```diff
@@ -56,6 +56,7 @@
     "mutant-pull-reduce": "^1.0.1",
     "open-external": "^0.1.1",
     "patch-context": "^2.0.0",
+    "patch-drafts": "0.0.5",
```
mixmix: :fire:  This needs to be published and updated before merging

PR #166, opened 2017-12-04
deps - patchbay-book@1.0.3

frankiebee:
patchbay-book@1.0.5 has not been published to npmjs so trying to `npm i` throws an error. 

This PR fixs version number for patchbay-book to current published version on npm


arj03:
Seems like @Happy0 fixed this in latest commit. Thanks nevertheless for the PR :+1:

PR #168, opened 2017-12-10
Bump patch-core and ssb-chess

Happy0:
Bumps to version of ssb-chess with piece capture differences and to patchcore 1.20.1.

Doing this as a pull request as I'm not sure if you're ready to move to 1.20.1 of patch-core or not :).


Happy0:
Closing as I think it's enough to just bump ssb-chess on its own, since ssb-chess uses the more up to date version of patchcore internally (in its own package.json)

https://github.com/ssbc/patchbay/commit/3b93347dfd26b44b7fe01a4e874dc6947d049aaa

Issue #106, opened 2017-04-19
bulk-require is a bad idea


dominictarr:
This issue is a note about my objections to bulk-require, I know I'm not currently the main maintainer, so I'm not gonna force anyone on this, but just gonna point out the problems as I see them, but I acknowledge this is not my decision.

First of all: how bulk-require bit me. I had been running v6, and then I checked out 7 and it just acted weird.
turns out, I still had a `./build` directory, which was not checked in, as a artifact of v6. But this was being included as part of bulk load, because it was part of `'!(node_modules|junk)/**/*.js'`.

In this case, it broke things, and was obvious something was wrong, but in a different situation it could have broken things in a way that wasn't obvious. It wasn't a security hole, but this is how security holes happen - it was an _unintended data flow_. I think it's much better to be completely explicit about what code is included or not.

For example, maybe later we need some scripts to do something in patchbay, and we write some scripts, maybe if you include these scripts they don't seem to do anything, but actually they introduce a security vulnerability. It doesn't look like a vulnerability when it's just run as a script, so it gets merged, but the merger forgets that it should be excluded from the bulk-require glob, because it doesn't break anything.

If it was explicit, then for the scripts directory to be added, there would have to be a commit that adds it, making it less likely it happens by accident, also this wouldn't have loaded the js in the build folder either.

I tried to change the negative glob into a positive glob that at least explicitly lists the directories to load code from, but [bulk-require](https://github.com/substack/bulk-require) doesn't even document the negative globs, so I tried a few things and it didn't work.

This is _not simple_. It's using magic not to fight dragons, or to give genuinely new capabilities, just for lazyness.

Read any fairytale, using magic for lazy reasons ends in tragedy.

The honest/explicit solution is to run bulkify manually, and check the result into git, nothing wrong with that. You always know what you are getting then.

/cc @ahdinosaur @mmckegg @mixmix

mixmix:
Open to trying more explicit mode.

There was that script in old patchbay which did some of this right? Guess
I'm keen if there's a way which doesn't require heaps of fiddling.

The original motivation for using magic was certainly to make refactoring
less of a nightmare.

evbogue:
I second @dominictarr on this, `bulk-require` is a bad idea. It also seems lazy to me.

evbogue:
[...]

PR #169, opened 2017-12-13
remove bulk-require

dominictarr:
closes: https://github.com/ssbc/patchbay/issues/106
also, same in patchcore: https://github.com/ssbc/patchcore/pull/51

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
   },
   "homepage": "https://github.com/ssbc/patchbay#readme",
   "dependencies": {
-    "bulk-require": "^1.0.1",
     "bulkify": "^1.4.2",
     "cross-script": "^1.0.5",
     "dataurl-": "^0.1.0",
@@ -100,3 +99,4 @@
     "standard": "^8.6.0"
   }
 }
+
--- a/main.js
+++ b/main.js
@@ -8,15 +8,92 @@ require('setimmediate')
 
 const patchbay = { 
   patchbay: {
-    about: bulk(__dirname, [ 'about/**/*.js' ]),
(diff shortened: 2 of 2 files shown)


dominictarr:
Also, while writing this, I figured out something about how depject is used in the topic/type naming style...
when a directory goes _four layers deep_ the last layer is a bunch of gives that is loaded as a `first` or `map`, for example: `./message/html/{decorate,layout,meta,render}/`

although, `./app/sync/initialise/` contains only `settings.js` but it doesn't return anything so it would also call the rest too (maybe some plugged in apps also have this?)

ahdinosaur:
> when a directory goes _four layers deep_ the last layer is a bunch of gives that is loaded as a `first` or `map`, for example: `./message/html/{decorate,layout,meta,render}/`

yeah! @mmckegg took this technique to the next level with [`modject`](https://github.com/mmckegg/modject), where it matters if the last layer is a file or a directory.

mixmix:
the initialise pattern is one I started exploring in ticktack. That it exists here is because I've started to port the idea into patchbay. 

Have been meaning to look at this @dominictarr , sorry it's taken time

mixmix:
looks good. Needs `bulk-require` removing then ready to merge

dominictarr:
fixed

mixmix:
cool, I'll review this and get this in in the next release

package.json
```diff
@@ -34,7 +34,6 @@
   },
   "homepage": "https://github.com/ssbc/patchbay#readme",
   "dependencies": {
-    "bulk-require": "^1.0.1",
```
mixmix: not to self (mix) remove bulkiy if it's not needed any more

PR #174, opened 2018-01-20
Add keyboard tab controls

reconbot:
Tab keyboard shortcuts like Chrome or Firefox

I couldn't prove that windows has a different `Window` menu, if I'm wrong I'm happy to put back the conditional.

Chrome and Firefox has their tab controls under `File` but that seems weird here. Also I'd love to bring in a new tab window but that's a completely new view and I have no idea what it might look like. Additionally I'd love to do an undo close tab but I'm not sure how we'd want to keep history.

tabcontrol

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,6 @@ Patchbay is built using [patchcore](https://github.com/ssbc/patchcore) + [depjec
 
 This makes in very easy to create say, a renderer for a new message type, or switch to a different method for choosing user names.
 
-
 ## Setup
 
 Libsodium has some build dependencies. On ubuntu systems the following might help:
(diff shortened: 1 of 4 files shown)


arj03:
First of great work!

I think it would be really great with a save tabs on close and restore feature. I sometimes close patchbay by accident and have to find all the stuff again. Maybe something like <click me> to restore tabs from last sessions? You can look at how things are stored related to filters or draft for how you might go about that.

Minor nitpick: Maybe add the shortcuts to the README.md?

arj03:
1) sync relates to synchronious functions. There a file naming pattern across patchcore, patchbay and related projects.

2) https://github.com/mixmix/patch-settings

3) Take a look in app/html/tabs.js. Seems there already is some history code that might be applicable to what you're doing here.

4) As far as I remember it is the object.

mixmix:
hey @reconbot this is really sweet, thanks for adding this!
It actually fixes a longstanding bug where right-clicking crashed patchbay.
We can now access some of the feature in the right click again, which I've been really missing ❤️ 

I merged your work in https://github.com/ssbc/patchbay/pull/180
I'm sorry I didn't see your PR early. Feel free to poke me in a PM on scuttlebutt any time. Also happy to give you crash course on depject sometime.

Re: more features - YES! that tabs saving one sounds sweet, and I'm also a big believer in small PRs, which is why I merged your work here, because tab preservation is it's own thang.

mixmix:
also - just sent you an invite to collaborate on this repo, so we can _jam_ on PRs more easily in the future :heart:

PR #176, opened 2018-02-08
Async router normalisation

mixmix:
OH THIS FEELS SO fucking good

changes the handling of routing to : 
- async normalize locations
- meaning we can pause and _fetch message_ before passing them to the router

In the case of th clicking links to messages in the middle of threads, I've change the behaviour so that we open the thread at it's ROOT, then scroll down to the particular message

I feel awesome.
This fixes that stupid REPLY button behaviour that forks threads in Patchbay
Also added a `app.sync.locationId` which centralises the making of a unique id from a given location, which is used by the tab system - this makes comparisons against ids and locations easy.

Still needs:
- probably some refactoring
- smoke testing 🔥 to see if it breaks anything
- check if it effects any plugins ? (will only if they mess with the `router.sync.normalize` or `router.sync.router`)
- ....


mixmix:
to continue being excited 🎺 🎶 

This is gonna open up possibilities for clicking on links and then going "oh, this is a fricken gathering, let me send you to a special view dog!"

mixmix:
TODO - rm the "message decorator: root"

arj03:
I'll be testing this today 🎺

mixmix:
hey @arj03 thanks for the feedback. Can you tell me where exactly you're clicking on a message and getting it to scroll down? 

places I've been testings : 
- keyboard shortcut (o or enter) when a message is selecter
- clicking `Reply`
- clicking the timestamp

arj03:
Sorry mix, that wasn't very specific. 

I just tried the options you mention again. Clicking reply got me this error:

TypeError: Cannot read property 'call' of undefined
    at selectChild (/home/chrx/dev/patchbay-workie/app/html/scroller.js:50:54)
    at HTMLDivElement.scroll (/home/chrx/dev/patchbay-workie/app/html/scroller.js:56:5)
    at locateKey (/home/chrx/dev/patchbay-workie/app/page/thread.js:103:28)

The timestamp does indeed now jump to the message, great! Wonder why it wasn't before.

Keyboard shortcut also seems to work, but I just got one time where it didn't. So something there is a bit flaky there.

mixmix:
thanks for those details I'll revisit this and let you know when I think it's fixed

mixmix:
how about now ?

arj03:
Hit another bug:

TypeError: Cannot read property 'scrollIntoViewIfNeeded' of null
    at locateKey (/home/chrx/dev/patchbay-workie/app/page/thread.js:107:14)

mixmix:
mmm, looks like we need to check for undefined AND null ... which is strange, but I'll do it

mixmix:
this is embarassing. If it doesn't work can you send me the id of the message you're getting errors with so I can try and reproduce. Also let know if you have a slow computer or something in case there's some race condition I'm not seeing.

# _one more time_

arj03:
Yes my machine is slow :-) Tested it this morning with no problems so far. Gonna leave it for the whole day and see if anything comes up. 

Btw. this is really great!

arj03:
Has been solid all day. Good for merge :trumpet:

mixmix:
thanks for all the help testing this @arj03 

I reckon it's gonna make gatherings and books way easier to display. We no longer have to use the thread view for everything, we can make a custom view for a book.

arj03:
Found something that is a bit wierd. In this thread:

%P+bgwaZM+VB1KfsvfubMyiSPFBRK9vmjKGvitAZz4xQ=.sha256

I wanted to respond to kas message (the 3rd last message in the thread) clicking reply bumps me to the first message in the thread?

mixmix:
Yeah felt now no longer gives you nested replies. Nested replies and forms
need to be actually built. This is a step back in a way, but it is also an
improvement - reply used to break threads in other clients.

I saw the scroll to top thing, I don't know why it does that, it should
scroll you to the message you just clicked on

PR #177, opened 2018-02-10
remove PMs + my posts from /notifications

mixmix:
I got sick of reading redundent messages in `/notifications`

This is a mini PR which filters out : 
- messages I authored
- messages that are private (which are already in `/private` or `/inbox` which I view as their own dedicated notification area)

I'm going to merge this straight off because it feels small, but happy to discuss. I think notifications need a whole feature built around them in the future, so expect this to be scrapped

cc @Happy0 @arj03 ... who else needs to know?


arj03:
Fine by me. It would be really nice if notifications and especially private would display a small number indicating are unread messages in there.

Issue #159, opened 2017-11-13
direct replies to messages fork the thread (but they probably shouldn't)


mmckegg:
I know that you've talked about this @mixmix but I just wanted to add this as an issue. 

I think an easy fix would be for replies to always copy the root of the message they are replying to. This is how it is done in patchwork: https://github.com/ssbc/patchwork/blob/master/modules/page/html/render/message.js#L71 (it's a bit of a mess in there to be honest!!)

@mixmix Would you be interested in a pull request to bring the behavior inline with patchwork?

There is definitely the case where you do want to fork a thread, though. We need to warm up that discussion again sometime!

mixmix:
Hey my intention is to bring this inline with some work done on ticktack.

Specifically, using the fork tag in conjunction with the root tag. Root
presence implies nested reply in same thread. Root absence implies hard
fork/ different topic and space/ new thread.

This was discussed in ssb in public and dealt happy with the pattern.

Happy for PRs to implement that

mixmix:
Thanks for the issue

PR #182, opened 2018-02-20
#159 - fix patchbay reply forking thread

mmckegg:
This PR fixes the dreaded reply forking bug.

Looks like the patchcore `message.obs.thread` module is a bit messed up, so I just did the root resolution manually in patchbay. I believe this is also how it is done in patchwork.

HURRAY! 🎈 🎉

### DIFF
--- a/app/page/thread.js
+++ b/app/page/thread.js
@@ -14,6 +14,7 @@ exports.needs = nest({
   'message.html.render': 'first',
   'message.async.name': 'first',
   'message.sync.unbox': 'first',
+  'message.sync.root':'first',
   'sbot.async.get': 'first',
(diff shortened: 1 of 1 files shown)


mixmix:
I just merged a fix for this today, my fix jumps you to the location of the
message in the context of its thread

I'll have a look at what you've done when I've got a chance to see if
there's anything I should merge in

mmckegg:
My fix lets you reply to a specific message without forking the thread. Your fix would just put it at the end of the thread (rather than "in reply to").

Maybe a mix of both could be good!

mixmix:
I want to make scuttle-post which is the way to nest, fork, post without
having to know how to do it under the hood.

PR #185, opened 2018-03-17
re-add user filter, add root-only filter

mixmix:
image

hey @arj03 here's me putting the user filter back in. It doesn't persist like the others.

I've also added a 'root' filter, which was a feature requested by a colleague - basically it lets you just review the heads of the threads to find things again a little more easily


mixmix:
TODO (note to self) extract the lightbox that SoapDog made for Ticktack and put the filter settings in there - that thing was _dope_

arj03:
The roots filter is nifty! Actually really nice after testing it. Nice! I don't mind the people filter.

PR #187, opened 2018-03-21
Use ssb-search instead of ssb-fulltext. Super fast!

arj03:
ssb-search uses flume and is much lighter overall


mixmix:
spinning this up now - once index is done will merge if it's working well!

mixmix:
OH MY GOD it's fast.

_MERGE_

mixmix:
one definite improvement here is that this searches titles as well as friggen body text -- the fulltext search just didn't work for some things.

Wow this is a joy to use. 

thanks @arj03 and @dominictarr   :heart: :heart: :heart: :heart:

mixmix:
hey @dominictarr / @arj03 looks like I'm getting streaming results which are looping.

eg. when I searched "blob viewer" and scroll down I see multiple repeats of the same headline. e.g. : 

---

# Community Gardening Technical Documentation

by happy0

dominictarr:
oh, hmm, I havn't tested it that way. sounds like the problem is that ssb-search stream doesn't support continuing from the previous stream in the same way as ssb-fulltext did, _or_ pull-more is looking at the wrong thing. Simplest fix is just making the `limit` be a much higher number.

arj03:
Hmm, can't reproduce this

PR #191, opened 2018-04-17
Add quote button

arj03:
This is mostly the work of https://github.com/gmarcos87. I had fun reviewing and grokking some more patchbay internals. Leaving it here in case others will find it useful.

### DIFF
--- a/main.js
+++ b/main.js
@@ -27,6 +27,12 @@ const patchbay = {
   }
 }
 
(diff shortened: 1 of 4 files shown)


arj03:
I'm gonna merge this as I find it quite useful, not sure if post thing is the best way, but I think that can be sorted out later.

PR #192, opened 2018-04-17
Encrypt blobs in private messages

arj03:
This is not ready yet as all the pieces are not in place yet. But this is enough so that blobs are encrypted. Missing part is showing the encrypted blobs using the unbox key.

Related: 
https://github.com/ssbc/patchcore/pull/60
https://github.com/ssbc/ssb-ref/pull/15


arj03:
With the latest commit this is now working :fireworks: Thanks @mmckegg!

cc @mixmix

PR #194, opened 2018-05-05
ssb-loomio / polls

mixmix:
This adds the first pass of polls to Patchbay


mixmix:
Also adds a new `app.html.modal` (from Ticktack!) which makes pop-ups super easy. See how it's used in `patchbay-poll`

PR #195, opened 2018-05-27
mvp image search

mixmix:
selection_615
selection_614
selection_613


mixmix:
this is a littlebit messy but the mess is contained !

PR #196, opened 2018-05-28
upgrade scuttlebot + ssb-private

mixmix:


mixmix:
hey @arj03 do you reckon this is stable enough to run in master?

If you think it is, please feel free to merge.

arj03:
Yeah, I've been running 11.3 for a while and its really stable for me now.

mixmix:
Woop!

There's a thing I'm watching where jumping between versions of softener
plugins sometimes breaks my indexes... Like my public feed decides to
break, or inbox won't load, but notifications will.
Only thing I can see is the counts in "sbot status" stall before getting
fully up to sync.

Anyway, just wanted to mention it in case you hit it. Wiping indexes works
fine to fix it

arj03:
What version of secure-scuttlebutt do you have? Dominic did some fixes to the indexing in 18.0.5.

PR #200, opened 2018-06-12
bump patch-inbox, standardise some styles

mixmix:
selection_639

This changes : 
- the "new" button on the `/inbox` page
- some styling across the app of buttons

The styles aren't amazing, but it's working!
If you don't like it, you can still use the composer in `/private`


main.js
```diff
@@ -28,8 +28,8 @@ const patchbay = {
 }
 
 const post = {
-  patchbay: {
-    message: bulk(__dirname, [ 'post-patchcore/message/**/*.js' ])
+  'patchbay-post': {
```
mixmix: hmm me changing this was a little superstitious .. I was having some bad problems which turned out to be me making a typo. Need to revert this

PR #202, opened 2018-06-16
styles remix

mixmix:
This is a start of adding what are basically CSS variables.
It's going to make theming or choosing your own color scheme super easy!

image

There's more I'm going to do, but I think it would be good to merge what I've done so far

### DIFF
--- a/app/html/app.mcss
+++ b/app/html/app.mcss
@@ -1,5 +1,5 @@
 App {
-  background-color: var(--app-border-background)
+  background: var(--app-border-background)
   position: absolute
 
   top: 0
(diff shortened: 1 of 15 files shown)


mixmix:
gatherings styles are going to need fixing after some changes I've made to how basic messages are rendered (making them much more concise) : 

image

arj03:
This is really lovely. Great work on the colors, the dark inbox fits much better now. The only thing that looked a bit strange was that the time for the post has moved. For very short messages it can be hard to see if the time is for the current message or the one following.

Anyway great work!

mixmix:
Yeah trying it different locations for things. Not sure i like it either.
Also, I broke the styling for gatherings cards, need to fix that.
I imagine I I'll iterate the default card layout. It's also something we
could add as a setting so people can choose their own!

PR #203, opened 2018-07-05
ssb-mutual fix

mixmix:
This isn't a fix I'm that happy with.

ideally ssb-mutual should be installed as an sbot plugin running in the background process.
I tried this but then you get some nasty recurssive explossion because this plugin attached `sbot` to the initialised class, which secret-stack then finds and tries to merge into the plugins. Repeat ad infinitum.

in the meantime ... this fixes a race condition


dan-mi-sun:
thanks for looking into this so quickly @mixmix

PR #204, opened 2018-07-06
scuttle-shell : ALL IN

mixmix:
This takes all sbot running responsibilities and hands them to [scuttle-shell](https://github.com/soapdog/scuttle-shell)

one advantage is that scuttle-shell is spun up as a detached process which can continue to run independent of the client that happens to be running.

What do you think?


mixmix:
hmm, this is working in ubuntu 16.04 (or maybe 17.04 ?)
not working at all in ubuntu 18.04 which I just installed ... not sure what the problem is, having a hard time debugging this D:
there's something killing systray immediately (and it's not sbot) ... an exit code 127 ("command not found") being emitted but don't know which command isn't found as there's no message!!!! ARG

cryptix:
Hey @mixmix! Sounds like what had quorralls with as well. See https://github.com/soapdog/scuttle-shell/issues/4

mixmix:
closing in favour of cryptix' new branch!

PR #208, opened 2018-07-14
Query yo

dangerousbeans:
was trying to test query view thing but couldn't get to run without these changes, looks awesome tho :3 (fixed branch)


mixmix:
thanks for this, sorry I didn't see it earlier. these changes did make it in but in a different commit.. somewhere

PR #209, opened 2018-07-16
bump ssb plugins

mixmix:
I noticed testing Tim's ngx-ssb-client that some of our plugins are crusty!
I bumped them all to hopefully scoop up some fixes and save re-indexing.

I've not review why ssb-ws is a major bump. I hope the minor bumps are all fine?
my plan is just to run this and see if breaks. Care to join me?


arj03:
@mixmix Great. I was going through the different plugins trying to fix some of the worst security holes. See https://github.com/ssbc/patchbay/pull/210. Have you had any problems yet?

arj03:
I merged the other one. I don't think ssb-ws is needed. It should only be used for web sockets, so I think this was for the lite browser client.

mixmix:
I had a sneaking suspicion ssb-ws was used for blob serving (or I heard someone say that ?)
Any idea @arj03 ?

arj03:
Right it does. Btw. scuttlebot has been running [ssb-ws 2.1.1](https://github.com/ssbc/scuttlebot/blob/master/package.json#L64) for quite a while so I don't think we should have any problems. But that probably also means that it's not really needed here :smiley:

PR #214, opened 2018-07-26
Add calendar to menu items

lancew:
Changes to make it so that /calendar is available in the menu far right.

This is really a request for comments, it seemed to work for me; But was not really sure I was doing things "right".

### DIFF
--- a/app/page/calendar.js
+++ b/app/page/calendar.js
@@ -1,18 +1,34 @@
 const nest = require('depnest')
 const { h, Array: MutantArray, map, Struct, computed, watch, throttle, resolve } = require('mutant')
+
 const pull = require('pull-stream')
 const { isMsg } = require('ssb-ref')
 
-exports.gives = nest('app.page.calendar')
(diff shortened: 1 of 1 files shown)


app/page/calendar.js
```diff
@@ -1,18 +1,34 @@
...
 
-exports.gives = nest('app.page.calendar')
+exports.gives = nest({
+  'app.page.calendar': true,
+  'app.html.menuItem': true
+})
 
 exports.needs = nest({
   'message.html.render': 'first',
+  'app.html.scroller': 'first',
```
mixmix: thanks for this @lancew 
This looks right, apart from link 14 - we don't need the app.html.scroller in this component, so remove that line then it's ready to merge. :pineapple: :boom:

lancew: Oh drat, yes I added that and then removed where I was using it and forgot to remove it here. Thanks

mixmix:
thanks @lancew , mergerd!

PR #216, opened 2018-08-06
Fix reply and quote in threads

arj03:
I've been wanting to fix quote not working in threads, and also reply since this [commit](https://github.com/ssbc/patchbay/commit/535b9a31c844497006ff5f0bf0dc391d3df69ae0) (I think). I'm posting this here to get some help on two issues:

- This introduces a new reply button so currently we have two. How do I remove the patchcore one?
- When I click quote it seems that goTo is called twice. That is why I added the setQuote instead using addQuote. This seems to be a general thing?


arj03:
Found an ugly fix for overwriting patchcore reply.

arj03:
The double goTO is because api.history.sync.push(loc) which calls api.history.obs.location()(loc => api.app.sync.goTo(loc || {})) in main.js. That is an observable on location ;-)

arj03:
Right! Now it is working and patchbay navigation seems much more smooth now that we don't have double/triple navigation. 

@mixmix I can haz review?

app/sync/goTo.js
```diff
@@ -35,10 +35,17 @@ exports.create = function (api) {
...
+
       if (page) {
         tabs.select(locationId)
         api.history.sync.push(loc)
 
+        if (loc.action == "quote" && page.firstChild && page.firstChild.addQuote) {
+          page.firstChild.addQuote(loc.value)
+          tabs.currentPage().scroll('last')
+        } else if (loc.action == "reply")
+          tabs.currentPage().scroll('last')
```
mixmix: if the thread is not already open the scroll will hit a race condition (I think) where the thread won't all be loaded, and so this action will cause a scroll to the last child currently rendered.... maybe

I guess the question is what is the context in which a person hits this quote or reply button.

There's also a thing I added you should know about where when a person goes to a message location, if that location is a message and it has a root, I tell the thread to render the root, and store the location as the root (because this means any other clicks with the same root will open that tab). Not sure it's relevant here but it's a funny little trick

post-patchcore/message/html/action/quote.js
```diff
@@ -11,7 +11,7 @@ exports.create = (api) => {
   return nest('message.html.action', function quote (msg) {
     return h('a', {
         href: '#',
-      'ev-click': () => api.app.sync.goTo({ action: 'quote', key: msg.key, value: msg.value })
+      'ev-click': (ev) => { ev.preventDefault(); api.app.sync.goTo({ action: 'quote', key: msg.key, value: msg.value }) }
```
mixmix: I might need to check how this behaves with `app.sync.locationId` (which makes unique ids for locations)

mixmix:
# Summary

Sorry the above were just thoughts, not requests.
I tested this out and here's how it behaved:
- :heavy_check_mark: quotes and reply button
- :x: closing tabs leaves no tab selected

The former behaviour was that as you click (or mouse-shortcut) between tabs, a history of where you last were builds up. When you close a tab, that tab is closed, and you're the tab you were on before that is brough into focus. At the moment no tag is brought into focus (and if you use a keyboard tab navigation shortcut it throws an error).

To be more concrete here's an example. I've got 3 tabs open A B C, and I've been clicking around. So that the history of where I've been (newest last) is like : 
```
history: [ A, B, C, A, C ]
```
So I'm on C, and I close it. As part of this action I strip the C from the history record (because we're not tracking it any more) and we're left with this: 

```
history: [ A, B, A ]
```

since we're also observing the current "location" (the last entry in the history), this triggers A to have goTo run on it.

mixmix:
I wrote up some thoughts on the whole mess of hypertabs + history here : %i2ZCCr82Jld4oVtUapAGz66dBNVvB0eYq8q4UpZbfqY=.sha256

mixmix:
hey @arj03 I've reverted a couple of changes which were breaking some behaviour (e.g. one piece of code suggested you should always go to the tab before the current one).

I've merged this because it gets the reply + quote fix in. We can mess with history / tabs more in another branch if you want. I think you mentioned something about performance at some point, but I've not experienced any problems and I not sure it was actually a problem for you?

PR #236, opened 2018-08-14
Fix keyboard shortcut scroll error

### LINKED ISSUE #222: TypeError: Failed to execute 'scroll' on 'Element'
I was trying to debug something and ran `Ctrl + Shift + K`, which triggered an error.

```
TypeError: Failed to execute 'scroll' on 'Element': parameter 1 ('options') is not an object.
    at genericShortcuts (/home/christianbundy/Source/patchbay/app/sync/catch-keyboard-shortcut.js:65:33)
    at root.addEventListener (/home/christianbundy/Source/patchbay/app/sync/catch-keyboard-shortcut.js:24:11)
```

christianbundy:
Resolves #222

### DIFF
--- a/app/sync/catch-keyboard-shortcut.js
+++ b/app/sync/catch-keyboard-shortcut.js
@@ -44,9 +44,15 @@ function textFieldShortcuts (ev) {
 }
 
 function genericShortcuts (ev, { tabs, search, goTo, back }) {
-  const scroll = tabs.currentPage().scroll || noop
+  let scroll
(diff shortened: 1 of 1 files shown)


mixmix:
there's another problem here where `scroll` appears to be an existing method on some dom elements, but defined by someone / something else and with a different API

so this will solve some of the issue but I think the method might need to be ultimately renamed to avoid conflicts (or the api moved to match?)

anyway, thanks for this

PR #238, opened 2018-08-14
Make quote work again

arj03:
This should fix so we don't get double quote and also make keyboard shortcuts behave the same way as clicking close. Depends on https://github.com/hyperhype/hypertabs/pull/14

### DIFF
--- a/app/html/app.js
+++ b/app/html/app.js
@@ -29,7 +29,8 @@ exports.create = function (api) {
     // runs all the functions in app/sync/initialise
 
     api.history.obs.location()(loc => {
-      api.app.sync.goTo(loc || {})
+      if (!api.app.html.tabs().currentPage()) // close
+        api.app.sync.goTo(loc || {})
     })
 
     return App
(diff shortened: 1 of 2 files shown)


arj03:
I have updated the PR with the latest hypertabs. This should be ready to rock.

christianbundy:
Hey @arj03!

Is there an issue for the "double quote" bug, or could you help me understand the problem this solves? I'd love to review but I'm unsure what to test (or what `// close` means in that context).

arj03:
@christianbundy sure

Go to a thread and hit the quote button on a message. Right now on master this inserts the message twice in the compose box. With this change it is only inserted once.

mixmix:
@arj03 have just spun this up and it breaks navigation for me :fire: 

Steps:
1. open a few tabs
2. navigate between them (using keyboard shortcuts or mouse)
3. close the current tab (using keyboard `x` or clicking the close)

image

expected outcome is for you to be taken to the last tab you were on, not be cast into the void between tabs D:

---

If this is being a total nightmare to solve, perhaps another path like right-click to "copy text as quote" could be much simpler.

PR #244, opened 2018-08-16
Sort menu alphabetically

christianbundy:
**This PR sorts the menu alphabetically, and removes the inline `order` styles.** I spent about an hour banging my head against the Mutant documentation when I realized my sorted array wasn't showing up correctly in the UI, but after removing the inline `order` CSS everything works great.

I'm still learning both depject and Mutant, so please let me know if there's a more idiomatic way to do this. I'm not sure whether `app/html/menu.js` is the best place for `sortByProp`, maybe it should go somewhere else? I'd appreciate suggestions and/or a review!

---

For future archaeology, I found https://github.com/mmckegg/mutant/issues/6 to be a great jumping off point for learning to sort observables with Mutant and Patchbay. I don't think I'd have the intuition to use `computed()`, so I'm hoping that was the right call.

### DIFF
--- a/app/html/menu.js
+++ b/app/html/menu.js
@@ -1,5 +1,5 @@
 const nest = require('depnest')
-const { h, Value, when, computed } = require('mutant')
+const { h, Value, when }  = require('mutant')
 
 exports.gives = nest('app.html.menu')
 
@@ -18,19 +18,9 @@ exports.create = function (api) {
     const hoverClass = Value('')
     const connectionClass = when(api.sbot.obs.connection, '', '-disconnected')
 
-    const sortByProp = (prop) => (a, b) => {
-      if (a[prop] < b[prop]) {
-        return -1
-      }
-      if (a[prop] > b[prop]) {
-        return 1
-      }
-      return 0
-    }
-
     const menuItems = api.app.html.menuItem(api.app.sync.goTo)
-    const sortedMenuItems = computed([menuItems], items =>
-      Object.values(items).sort(sortByProp('text'))
+    const sortedMenuItems = Object.values(menuItems).sort((a, b) =>
+      a.text.localeCompare(b.text)
     )
 
     // TODO: move goTo out into each menuItem
--- a/app/html/menu.js
+++ b/app/html/menu.js
@@ -1,5 +1,5 @@
 const nest = require('depnest')
-const { h, Value, when }  = require('mutant')
+const { h, Value, when } = require('mutant')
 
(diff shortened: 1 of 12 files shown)


app/html/menu.js
```diff
@@ -18,13 +18,28 @@ exports.create = function (api) {
...
     const connectionClass = when(api.sbot.obs.connection, '', '-disconnected')
 
+    const sortByProp = (prop) => (a, b) => {
+      if (a[prop] < b[prop]) {
+        return -1
+      }
+      if (a[prop] > b[prop]) {
+        return 1
+      }
+      return 0
```
mixmix: i think you could just do this: 

```js
 const sortByProp = (prop) => (a, b) => a[prop] - b[prop]
```

app/html/menu.js
```diff
@@ -18,13 +18,28 @@ exports.create = function (api) {
...
+    }
+
+    const menuItems = api.app.html.menuItem(api.app.sync.goTo)
+    const sortedMenuItems = computed([menuItems], items =>
```
mixmix: can you try this without `computed`, I don't thin menuItems is an observeable (as in it's not a value that's been set up to dynamically change)

mixmix:
couple of minor tweaks then ready to go I think.
This is a good idea, that menu is a mess (TBH I use keyboard shortcuts to get around and that sidesteps this uugly menu!)

mixmix:
I've added you as an owner of the patchbay repo so you can push branches up - makes collaborating easier

christianbundy:
Thanks for taking the time to review @mixmix! Here's a quick summary of the changes:

- I wasn't able to use the `a[prop] - b[prop]` comparator because it returns `NaN` on strings. I was able to find [`String.prototype.localeCompare()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) which seems to do what we want.
- I removed `computed()` and nothing of value was lost, thanks for pointing that out!
- Some menu items were getting `order: $n` styles applied by dependencies, so I've removed `item.style.order` properties from all menu items by default. I *think* this was the right move, but please let me know if there's a more elegant way to do this. 

I need to look into the keyboard shortcuts! I *just* learned about the Vim-like keybindings, but I'm sure there's other stuff I'd love to use.

mixmix:
MERGED !!

would be sweet to have a page which was `/shortcuts` and had a copy of the shortcuts that I think I listed in the README

PR #246, opened 2018-08-17
Post rank

arj03:
First version


app/html/search-bar.js
```diff
@@ -74,7 +74,7 @@ exports.create = function (api) {
     function getPagesSuggestions (word) {
       const pages = [
         'blogs', 'calendar', 'posts', 'public', 'private', 'inbox', 'profile', 'notifications', 'settings',
-        'gatherings', 'chess', 'books', 'imageSearch', 'polls', 'query', 'dark-crystal'
+        'gatherings', 'chess', 'books', 'imageSearch', 'polls', 'query', 'dark-crystal', 'postRank'
```
mixmix: NOTE TO SELF : oh god this thing is out of control, replace this with depject already!

app/page/postRank.js
```diff
@@ -0,0 +1,184 @@
...
+
+exports.create = function (api) {
+  return nest({
+    'app.html.menuItem': menuItem,
+    'app.page.postRank': page
+  })
+
+  function menuItem () {
+    return h('a', {
+      style: { order: 1 },
```
mixmix: I'm merging this but know that christianbundy just went through removing all the order over-rides so he could sort them alphabetically

app/page/postRank.js
```diff
@@ -0,0 +1,184 @@
...
+
+    console.log(myChannels)
+    console.log(imFollowing)
+
+    var messages = {}
+    
+    console.log("pulling", new Date())
+
+    return pull(
+      api.sbot.pull.log({ reverse: true, limit: 10000 }),
```
mixmix: This is fine for a first pass.


I would _highly_ recommend using ssb-query, and fetching just the `post`  and `vote` messages and merging with `pull-merge`.

Can also add a next-stepper I built for ssb-query called `pull-next-query` which means it will only load 100 messages at a time rather than a full 10000 into the UI.

If you want any help setting that stuff up let me know

arj03: Thanks!

This was mostly just porting over some existing code I wrote a while ago for fun. I've added the missing mcss. Forgot to check that in.

mixmix:
@arj03 I've merged this, left a couple of comments which might be useful for further iterations.

I've added `standard` to patchbay to help avoid minor bugs, prune dead code, and have a rough shared style across the project. 
Can you please run `npm run lint` on this and tidy up the unused variables at some point? That command will autofix what it can (like white space, etc)

PR #255, opened 2018-09-05
Don't init suggestion cache for abouts during startup

arj03:
This decreases the startup time for me from 18 sec to 12 sec. The suggestions will just be loading when we need then instead so we shouldn't loose any functionality (I think?). 

The test was conducted this: https://github.com/ssbc/patchcore/pull/69

### DIFF
--- a/app/sync/initialise/suggestionCaches.js
+++ b/app/sync/initialise/suggestionCaches.js
@@ -11,7 +11,8 @@ exports.create = function (api) {
   return nest('app.sync.initialise', init)
 
   function init () {
-    api.about.async.suggest()
+    // lazy load abouts on first use, can be quite heavy during startup
+    //api.about.async.suggest()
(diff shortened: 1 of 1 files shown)


christianbundy:
Do you have any suggestions for how to best test this? I'm testing your branch but it takes 5 seconds from `electron index.js` to `/notifications` loading in full, both on your branch and `master`.

Also, do you think it may be worth just deleting the code rather than commenting it out? We'll always have the previous code in our history, and the reason for the change should remain in the blame.

arj03:
@christianbundy it must be in combination with the patchcore patch then.

I think for these kind of things it would be good with an idle concept where we load / crunch stuff. So first init the app as quickly as possible and then start things of an idle stack when nothing is going on. I'm a bit wary of using setTimeout or similar for these kind of things as that is known to be really bad for batteri life.

PR #256, opened 2018-09-06
Replace /public with /posts in default tabs

christianbundy:
This PR means that future versions will default to opening `/posts` rather than `/public` when the application is opened with no "default tabs" customization. This will not effect users who have customized this to their liking.

It's worth mentioning that this is likely a personal preference, so if it's not a good candidate for `master` then I don't mind closing this at all. Also, if there are any unresolved bugs in `/posts` I'd be happy to take a look at them before merging is considered.

### DIFF
--- a/app/sync/initialise/settings.js
+++ b/app/sync/initialise/settings.js
@@ -12,7 +12,7 @@ exports.needs = nest({
 
 const defaults = {
   patchbay: {
-    defaultTabs: ['/public', '/inbox', '/notifications'],
+    defaultTabs: ['/posts', '/inbox', '/notifications'],
     accessibility: {
       invert: false,
       saturation: 100,


mixmix:
hey @christianbundy can you please push your branches to this repo.
If I want to tweak anything having to pull your code from your repo is mildly annoying :heart: 

Thanks for this change, it's been on my mind too

christianbundy:
Yeah, I've been using the `@user/branch` style from [git-ssb-intro](https://github.com/noffle/git-ssb-intro#alternative-collaboration-without-forks) to prevent naming collisions without forking, but all of my branches are being pushed the the [ssbc repo](https://github.com/ssbc/patchbay/network). I'll stop doing that if it's confusing, maybe that's only useful for `@christianbundy/master`?

Sorry for the trouble, thanks for pointing out the pain point!

mixmix:
Oh that's good to know. Sorry i mistook that notation for
christianbundy:branch
Yeah i think it's overkill, it pretty obvious who is authoring a branch /
pr imo

PR #260, opened 2018-09-20
scuttle-shell

cryptix:
uses scuttle-shell@0.1 which is based on sbot@12

@mixmix this was based on your old branch. I'm not sure about two of the deps in the pkg.json I dragged back in (ssb-horcrux and style-resolve).


background-process.js
```diff
@@ -1,34 +1,28 @@
...
-  .use(require('ssb-friends'))
-  .use(require('ssb-meme'))
-  .use(require('ssb-private'))
-  .use(require('ssb-query'))
-  .use(require('ssb-search'))
-  .use(require('ssb-unread'))
-  .use(require('ssb-ws'))
-  // .use(require('ssb-mutual')) // this is has recursion problems atm
+// Check if scuttle-shell is already running
+// TODO - make this check for scuttle-shell specifically (and not just an sbot)
```
cryptix: @mixmix I'm also not sure about this comment.

How would the shell be different from another sbot with the same plugins?

christianbundy:
Is this branch still the best way to try scuttle-shell?

mixmix:
yes but i've not tried it recently.

cryptix:
I’m using it all the time and rebase it when I see something new intersting. ssb-about@2 should already work for instance.

cryptix:
superseeded by #295

PR #262, opened 2018-10-03
add ssb-blob-files (to replace blob.html.input)

mixmix:
this is a courtesy PR

y'all should know I've extracted and refactored a bunch of the blob creating logic.
I plan to also open a PR upstream in patchcore.
I'd like to make [**ssb-blob-files**](https://github.com/ssbc/ssb-blob-files) the thing we all maintain and add to


arj03:
Thank you 🙏 

Just out of curiosity, now that patchcore is no longer used in patchwork (at least in its current form) does it make sense to put more energy into it? It seems like the way forward is these smaller modules that you are creating.

mixmix:
not really, but I want to go and rip out all instances of that old input code so that no-one can accidentally keep watering it

PR #263, opened 2018-10-04
[WIP] Use variables for all colors

christianbundy:
Hi! I'm working on some themes and wanted to see whether this would be useful upstream. My goal is to implement a [base16](https://github.com/chriskempson/base16)-like theming system, so that any theme (just a collection of colors) could be dropped into the app. This branch currently uses [solarized light](http://www.zovirl.com/2011/07/22/solarized_cheat_sheet/), but it supports other themes too.

Current problems:

- ~~`app/styles/css/global.js` doesn't seem to be respecting variables~~
  - ~~specifically `body { color: var(--primary); }`~~
- ~~some colors are being set dynamically in JavaScript~~
- ~~the lightbox has an opacity issue where the background-color doesn't seem to be working~~
- probably a bunch more, but I've been using this as my daily driver and it works fine

No problem if this isn't useful, I'm happy to close, just wanted to offer the contrib if it was something others were into. :beers:

### DIFF
--- a/about/html/edit.mcss
+++ b/about/html/edit.mcss
@@ -19,7 +19,7 @@ AboutEditor {
 
           width: 100%
           height: 100%
-          background-color: rgba(0, 0, 0, 0.5)
+          background-color: var(--highlight)
         }
       }
     }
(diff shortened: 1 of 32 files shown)


HendrikPetertje:
Sorry to ask, but what happened here? is there a new branch, pull request or baked in feature that continues on what you started here? (I like it a lot!)

christianbundy:
@HendrikPetertje This worked as a prototype, but it came with a large theme change which I don't think was necessary. I'll likely coordinate with @mixmix and see about redoing this, making sure that we can stick with the default Patchbay theme rather than switching over to Solarized.

PR #264, opened 2018-10-11
Don't use socket on Windows

arj03:
@mixmix can you test this on Windows?

### DIFF
--- a/config.js
+++ b/config.js
@@ -20,13 +20,17 @@ exports.create = (api) => {
       const keys = ssbKeys.loadOrCreateSync(Path.join(config.path, 'secret'))
       const pubkey = keys.id.slice(1).replace(`.${keys.curve}`, '')
 
-      config = merge(config, {
-        connections: {
-          incoming: { unix: [{ 'scope': 'local', 'transform': 'noauth' }] }
-        },
-        keys,
-        remote: `unix:${Path.join(config.path, 'socket')}:~noauth:${pubkey}`
-      })
+      if (process.platform !== "win32") {
+        config = merge(config, {
+          connections: {
+            incoming: { unix: [{ 'scope': 'local', 'transform': 'noauth' }] }
+          },
+          keys,
+          remote: `unix:${Path.join(config.path, 'socket')}:~noauth:${pubkey}`
+        })
+      } else {
+        config = merge(config, { keys })
+      }
     }
     return config
   })


cryptix:
hmmm.. when w me started this endeavor initially, I thought windows supports these. But looking at it again, it looks like the support for this is still only win10 and upwards? [src](https://blogs.msdn.microsoft.com/commandline/2017/12/19/af_unix-comes-to-windows/)

I wonder if we can probe for this feature instead of making this platform wise. I can try this on win7 and 10 VMs next week. Already started setting them up for scuttle-shell testing.

arj03:
Yep that would be great @cryptix. This patch was just to bring it in line with 7.14 so that @mixmix can get the release out of the door.

mixmix:
I've got win 10., can test soon. Dio th think we need to support 7

christianbundy:
FWIW, not even Microsoft wants to support it -- Windows 7 lost mainstream support in January 2015 and Windows 8 lost mainstream support in January 2018, so I think targeting Windows 10 only is probably a reasonable move.

config.js
```diff
@@ -20,13 +20,20 @@ exports.create = (api) => {
...
+          connections: {
+            incoming: { unix: [{ 'scope': 'local', 'transform': 'noauth' }] }
+          },
+          keys,
+          remote: `unix:${Path.join(config.path, 'socket')}:~noauth:${pubkey}`
+        })
+      } else {
+        config = merge(config, {
+          keys,
+          remote: `net:127.0.0.1:${config.port}~shs:${pubkey}`
```
christianbundy: If we want to be IPv-neutral, we could also do:

```suggestion
          remote: `net:localhost:${config.port}~shs:${pubkey}`
```

:man_shrugging:

arj03: @christianbundy see this comment in patchwork: https://github.com/ssbc/patchwork/blob/9433261107177d6178c2858447046b7b11991afc/index.js#L185

arj03:
@mixmix did you manage to create installers for Windows anyway without this patch?

config.js
christianbundy: 😱 TIL Windows handles localhost differently thani remember, thanks!

arj03: Exactly :D

arj03:
No worries. Here is a panda hug for you
[hug](https://blog.usac.edu/wp-content/uploads/2015/04/hug-a-panda-e1429292142334.jpg)

PR #278, opened 2018-11-08
fix reopening main window on macOS

mmckegg:
In the current master of patchbay, you can close the main window on macOS and Patchbay continues to run in the background. However, when you click on the dock icon, nothing happens. The only way to show the main window again is to right click, quit, and then relaunch.

**This PR adds an `on('activate')` handler that shows the window again.**

This is safe to merge as it mostly just brings the patchbay `index.js` inline with the patchwork one.

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,6 +21,13 @@ electron.app.on('ready', () => {
     quitting = true
   })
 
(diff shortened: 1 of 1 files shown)


mmckegg:
@mixmix says this is good!

PR #280, opened 2018-11-08
dont add sockets for windows at the moment

mixmix:
not sure who added this sockets stuff but it's breaking the windows installers D:

This is a temp fix which should work
cc @cryptix @arj03 @christianbundy


mixmix:
the way I did this PR broke things. I fixed it in master

PR #285, opened 2018-11-25
Add CSS for syntax highlighting from highlight.js

christianbundy:
## Before

```
const x = 42;
const f = a => x

f() === f(x) === 42

if (true) {
  // it's happening
}
```

## After

```js
const x = 42;
const f = a => x

f() === f(x) === 42

if (true) {
  // it's happening
}
```

### DIFF
--- a/app/styles/css/global.js
+++ b/app/styles/css/global.js
@@ -1,5 +1,10 @@
 const nest = require('depnest')
 const { assign } = Object
+const fs = require('fs')
+const path = require('path')
+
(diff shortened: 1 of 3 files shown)


mixmix:
THANKS FOR THIS !

I took core parts of your code and did something slightly cleaner here : https://github.com/ssbc/patchbay/blob/master/app/styles/css/highlight.js

your code made it super easy to add this though

christianbundy:
Looks great, thanks for taking the time to refactor! That looks very clean.

PR #287, opened 2018-12-04
Handle case where about.slice is not a function.

pietgeursen:
Patchbay just blew up for me and this fixes it. Not sure if the problem originates in ssb-about and that would make this a hacky fix...

@mixmix


pietgeursen:
Oh think you just fixed this. Closing...

mixmix:
yeah we need schemas instead of stupid if checkers ...

PR #288, opened 2018-12-04
Work with sbot 13.2.1

arj03:
Here you go @mixmix 

As cryptix pointed out, it is [this](https://github.com/ssbc/scuttlebot/issues/577) issue.

### DIFF
--- a/config.js
+++ b/config.js
@@ -30,7 +30,7 @@ exports.create = (api) => {
 function Connections (config) {
   const connections = (process.platform === 'win32')
     ? undefined // this seems wrong?
-    : { incoming: { unix: [{ 'scope': 'local', 'transform': 'noauth' }] } }
+    : { incoming: { unix: [{ 'scope': 'local', 'transform': 'noauth', server: 'true' }] } }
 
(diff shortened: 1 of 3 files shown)


christianbundy:
I'm not familiar with that config, are there any docs on this change so I can familiarize myself?

mixmix:
Thanks for sorting that so quickly.
Im also confused as to where to find docs. I think i read them once but
they didn't really detail all the options last looked

Is it worth upgrading the installers I just made if it takes me say an
hour?

arj03:
@christianbundy that is the whole problem. There is no documentation update for the new stuff (13.1). See the error I linked to ;-) Things should be documented in ssb-config.

Sbot 13.1 is a dominics rework of scope. 13.2 just adds the disable-links that doesn't do anything for patchbay, as it is not using that plugin anyway. And 13.2.1 is a minor bugfix for people hacking their own config and plugins.install blowing up their config. So in short, I wouldn't bother with a new installer @mixmix.

mixmix:
Good news is ssb- config now has more docs!

PR #291, opened 2018-12-04
Use `require()` instead of invisible Electron window

christianbundy:
I think this is an improvement? It's less code and *feels* cleaner, but I don't know if it's actually a very big performance improvement.

- `server.js` runs in the main process rather than opening an invisible Electron window
- the "server-started" event is replaced with a simple callback
- errors from the renderer are still passed to the terminal
- `console.log`s from the renderer only show up in the dev tools, not the terminal
- errors in the main process will be output as text (no more "A JavaScript error occurred in the main process" windows)

### DIFF
--- a/assets/base.html
+++ b/assets/base.html
@@ -2,27 +2,13 @@
 <html>
   <head></head>
   <body>
-    <script>
-      // redirect console to main process
-      const { log: remoteLog, error: remoteErr } = require('console')
-      const { log: localLog, error: localErr } = console
-
-      console.log = function (...args) {
-        localLog(...args)
-        remoteLog(...args)
-      }
-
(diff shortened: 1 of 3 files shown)


cryptix:
I think the main reason for doing it is having `sbot server` on a different OS process. Otherwise it could block the view code as far as I can tell.

christianbundy:
That makes sense! I'll leave it as-is then, thanks.

christianbundy:
I'm not sure that this is the right way to go about it, but I'm running into an issue where this is causing `window` to be defined in the server and a dependency is defaulting to a browser implementation that doesn't work.

Ref: https://github.com/pfrazee/parse-dat-url/issues/1

PR #294, opened 2018-12-13
Bump to version of patchcore which changes the behaviour when hovering over avatars

Happy0:
Bump to version of patchcore which changes the behaviour when hovering over avatars to display the user's display name as well as their ID.

Note: this removes the about/html/link depject module since this is
exposed by patchcore (with the new behaviour.)

https://github.com/ssbc/patchcore/pull/76 (commit f3613c3af31488ec38f9cdd73d0353e7d9a7a559)


Happy0:
haha, thanks. nice emoticon :D

mixmix:
@Happy0 love this behaviour. There was one regression, which was that this added `@` in front of all the names! So I've gone and changed that in patchcore so the behaviour is what it used to be in patchbay.

I made a mess of the git commits in the patchcore, don't look ):

Happy0:
@mixmix sorry about that! I actually wondered if I'd done that... I was like 'was that always like that? ... Well, mix did make some styling changes recently... ' I should have listened to my paranoia, haha :p.

Thanks for sorting it out :)

PR #300, opened 2019-01-14
Invert colors for main posts.

### LINKED ISSUE #293: Color of link text makes links unreadable in /posts view
When I upgraded to patchbay-linux-7.15.6-x86_64.AppImage, I could no longer read the links in /posts view, as there is not enough contrast between the purple link color and the black background. 
contrast

I _think_ that I have a default setup, I haven't made any settings changes since the update:

settings

justinabrahms:
re %EAKEBPJtZh/JJnI/ZS5vvZSW+3eEUUjOCDB827TvIy0=.sha256

Fixes #293 

## Before
screenshot from 2019-01-13 21-26-57

## After
screenshot from 2019-01-13 21-27-16

### DIFF
--- a/app/page/posts.mcss
+++ b/app/page/posts.mcss
@@ -110,8 +110,8 @@ ThreadCard {
 
     div.root {
       div.Markdown {
-        background-color: #000
-        color: #fff
+        background-color: #fff
+        color: #111
         line-height: 1.4rem
         padding: 2rem
 
--- a/app/page/posts.mcss
+++ b/app/page/posts.mcss
@@ -178,7 +178,7 @@ ThreadCard {
     section.content-preview {
       div.root {
         div.Markdown {
-          background: #6f0055
+          background: #ecdde8
           transition: background .2s ease-in
         }
       }


christianbundy:
I'm personally +1 on this, what do you think @mixmix?

One suggestion might be to add a border above `.content-preview` to illustrate that those aren't part of the original post.

mixmix:
Can you just make this make the not-selected boxes the same colour (blackish), but ensure the `a` tags are white.

That would be an incremental change that would solve the problem. Changing the backgroung to white is something that could be sweet to do, but needs more design - as Christian points out just dropping it to white but not leaving any demarcation between the root message content and the preview of responses makes this a visual mess (IMO). So let's do that in a different PR

hardfire:
+1, I had this override as a personal css with rounded corners of user's images.

justinabrahms:
I think the intent of this change is "remove the black background" rather than "fix the link colors". I'd like it if when you inverted the colors, you actually got an inverted color space (either everything dark or everything light). To that end, I'm happy to change the color of the post to be a different light shade, but probably won't update this to just be a less dark dark shade.

If that's the thing you want.. that's legit. I can just move these things to my personal css overrides. :smile:  I'll experiment with some different light colors tomorrow.

PR #303, opened 2019-01-16
Use ssb-friend-pub

arj03:
From the scuttlebot [thread](%PyiK5ksJlFD+cZCj5OZo7o7OdZxf6LaopOiwK75Ib1s=.sha256) a week ago:

`
Recently I have been working on ssb-friend-pub github git-ssb. The motivation was kawaiipunks Criticisms of Scuttlebutt thread. This is a plugin for sbot that allows an owner of a pub to assert ownership and providing methods to see which pubs are run by your friends or friend of friends. 
`

My understanding is that you want to get ssb-server into a stable state and then do a release of that @mixmix? So it might not be the best time to merge something like this. It should be relatively harmless by default.

I have been running sbot with pubs run by friends for the past 3 weeks and havn't had any trouble with messages not getting through.


arj03:
Thanks for taking your time mix. Really appreciate it!

I'll fix the technical things, sounds reasonable. The restart/reload is because we need to disable the global gossiping if enabled. I'll see if we can do without that.

About the scuttle shell integration then I'm really interested in hearing what @cryptix has to say. This is similar to a "tor only feature" I have already written in the same style, but currently doesn't work because electron for some reason does not allow proxy connections on the local machine. Probably security related but I can't seem to find anything about it, and if it can be disabled or opened for a specific port. So first question: scuttle-shell is not running electron right? If so, that is a big win right there.

The second point is that I'd be great to have a UI to configure some of these things. Right now I have them in patchbay settings. So I'm interested if one would rather but all of these things in scuttle-shell and if so what kind of UI does it have? Or should I rather create a minimal program that is only used for configuring these things based on patchbay, that somehow would be easily launched from scuttle-shell?

mixmix:
Restart relay is fine imo, we just need to tell the user that clearly
(maybe you did already)

Let's think about scuttle shell later on I guess. In particular I'm mindful
that cryptix needs a break from it and I don't want to put more more on him
to solve right now.

cryptix:
<3 yea, thanks mix. to be honest: Any UI apart from the systray thingy is totally out of scope for me. The library is just to limited and needs a new pick, like libui that Tim suggested but somebody else needs to start that...

arj03:
Really sorry if I said something that comes of the wrong way related to scuttle-shell. I havn't been following the discussions because I have to limit myself. I just got a bit excited because I saw it as a way to solve the tor problem.

arj03:
I'd like this merged, as its just after a release. Any objections, I should have fixed all the cases raised and the diff looks quite clean to me :)

christianbundy:
Just a quick question: when this says "only connect to pubs within X hops" does it really mean "only connect to *peers* within X hops"? I'm directly connected to a handful of people that aren't pubs, and I'm curious whether this would change that.

arj03:
@christianbundy it should say: "Only connect to pubs run by a peer within a certain number of hops". I'm open to other wordings, but for me that seems clear enough?

What do you mean by connected to people directly? Do you use seeds in your .ssb/config? Or do they have private ssb-device-addresses? This does not overwrite seeds, it only disables public gossipping, meaning "random" pubs.

arj03:
Ho ho ho, upgrading to socks 2.2.2 makes tor work inside electron and 2.2.3 makes it play better with latest version of tor. :cocktail: :dancer: :tada:

christianbundy:
> What do you mean by connected to people directly?

For example, I see that I often connect to @kas and others who use cjdns, but in my mind those aren't pubs (maybe they are?). Anyway, not a show-stopper, please continue. :upside_down_face: 

Regarding Electron, does this mean that we can just spin up Tor inside Patchbay *without* any host daemon? If so, that's a game-changer.

arj03:
Oh, that works because cjdns makes it behave like they are on the same lan right? This shouldn't change how that works.

Sadly you still need tor running outside, but it is enough simple to apt-get install it.

cryptix:
> but in my mind those aren't pubs

_pubs_ never had a really special role. It's just that the other peers never had addresses.

PR #306, opened 2019-01-17
Add SSB URI support

christianbundy:
This uses the ssb-uri module in both the search box and the link handler
to support the use of `ssb:` URIs. This does *not* change the title of
the tab, the sigil link presented in the search box, or any other
details of how sigil links are handled.

---

Oh, and this also switches us from the legacy URL parser (`require('url')`) to the "new" WHATWG API, which ships with JavaScript. More details [here](https://nodejs.org/api/url.html#url_url_strings_and_url_objects). I don't think that this has any downsides.

---

Check out <ssb:message:sha256:BRf34OH0jSTOZ6g5x6OdvodI0etfbs08mvaIYy2Sae4=> to test.


arj03:
Code looks fine to me. Is the idea that we can introduce more handlers, like dat: in the future?

app/async/catch-link-click.js
```diff
@@ -32,7 +31,15 @@ exports.create = function (api) {
       var href = anchor.getAttribute('href')
       if (!href || href === '#') return
 
-      var url = Url.parse(href)
+      var url
+
+      try {
+        var url = new URL(href)
```
mixmix: should be just this I think
```js
url = new URL(href)
```

app/html/search-bar.js
```diff
@@ -18,9 +18,11 @@ exports.create = function (api) {
     if (_search) return _search
 
     function goToLocation (location, ev) {
+      const prefixes = ['@', '#', '%', '&', '/', 'ssb:']
+
       if (location[0] === '?') {
         location = { page: 'search', query: location.substring(1) }
-      } else if (!['@', '#', '%', '&', '/'].includes(location[0])) {
+      } else if (prefixes.every(p => !location.startsWith(p))) {
```
mixmix: nice refactor!

mixmix:
Haven't tested this in live app but looks good to me!

there's a minor change I mentioned about that you can fix in master if you think it's worth it

mixmix:
p.s. great work @christianbundy , so good to have some standardised interfaces being supported into being by you :seedling:

app/async/catch-link-click.js
christianbundy: I think I had to add this because it was throwing errors on invalid URIs. Unfortunately `new URL()` doesn't accept a callback to handle errors, so you *have* to `try {} catch {}` unless you're sure it's always a URI.

christianbundy:
Thanks for the kind words and the quick review :heart: please ping me if this causes any trouble, I've been using it as my daily driver but I know I can't always catch *all* the edge-cases.

PR #316, opened 2019-02-13
[WIP] initial sync + change config before launch (ahoy)

mixmix:
selection_273

SUPER UGLY ... lots to improve!!
**DO NOT MERGE**


config.js
```diff
@@ -18,8 +18,11 @@ exports.create = (api) => {
 
     config = addSockets(config)
     config = fixLocalhost(config)
-    // config = pubHopSettings(config)
-    // config = torOnly(config)
+
+    if (false) {  // TODO figure out how to check if this is being run in/ out of electron
```
arj03: @mixmix it might be a good idea if these options we're used in ahoy as well, at least if that is running before patchbay. Let me know if you want to have a chat about it. Ahoy looks like a nice abstraction, so keen to help out however I can.

Happy0:
Just want to cheer lead / give some appreciation from the sidelines. This will be super valuable :D

mixmix:
This is ready to merge. I think it's stable. The bad news is I lost some UI work which made this look nicer. Happily, I took screenshots so will re-implement stuff in `ssb-ahoy`

**what I'd like reviewed** - I'm making breaking changes to how patchbay core files are layed out here ... I want to know if the renames feel ok, and whether people are ok for me to make these changes. I don't think this is going to be the end point, but I want to take this step to better understand the territory. Worst case if this is a poor idea we can revert what we don't like.

To test this is working on launch Patchbay with `$ AHOY=true npm start`

Requesting review / consent : @arj03 @Happy0 

I need to do a final test that this builds installers ok before we go live on this

arj03:
@mixmix I'm going to give this a spin and a proper code review over the weekend.

Kindly requesting a review of https://github.com/ssbc/patchbay/pull/326 :-)

arj03:
First observation: I don't see any peers under connections.web in the ahoy welcome screen. Could it be because I have friend-pub enabled? I seem to notice that the log from friend-pub only comes in after the initial screen (meaning before the indexing), but only the hosts, not connections.

config.js
```diff
@@ -18,8 +18,16 @@ exports.create = (api) => {
...
-    config = torOnly(config)
+
+    try {
+      window
+      // HACK (mix): patch-settings currently uses localStorage as persistence
+      // this means trying to load config outside context of electron crashes app ):
+      config = pubHopSettings(config)
+      config = torOnly(config)
+    } catch (e) {
+      console.log('TODO: decouple patchbay config from localStorage')
```
arj03: Seems like I'm hitting this one

index.js
```diff
@@ -1,170 +1,46 @@
...
-    },
-    { type: 'separator' },
-    { role: 'front' }
-  ]
-
-  Menu.setApplicationMenu(Menu.buildFromTemplate(menu))
-}
+    config,
+    plugins,
+    // appDir: '../patchbay', // uncomment when ssb-ahoy is symlinked in!
```
arj03: does this need to change now, or is the comment still valid?

arj03:
Added some comments in code review, probably the problem I'm seeing. 

Just wanted to note that this looks really great, the font can probably be bumped a bit. It is going to be really great for new users, excellent work!

mixmix:
@arj03 I lost some of the UI work I did (I slipped up with git, which hasn't happened in ages). Here's what the UI is going to look like once I re-implement it ! (I took a sceenshot luckily)

image

config.js
mixmix: mmmm I think I might need to finally improve `patch-settings` to use actual config files instead of `localStorage`

mixmix:
Selection_334

the unstyled "config picker" looks like this right now. At the moment you can only search `ssb_appname` that have already been set up. It's gonna be trivial to create new ones with what I've build though. (so in the screenshot `narm` implies i have `~/.narm` which has its own secret and config)

arj03:
Really sorry you lost all that styling :( That looked amazing!

Going to run through this real quick as I'm quite tired, but will get back to you on the structure.

I remember before one had to run a special command and its always enabled, right? I'm going to go crazy if I have to click next every time I start patchbay ;-)

Anyway, I fired it up and it seems to be working. I actually have a bunch of folder that I move around for testing, so this is going to be great for that.

Also the indexing screen is no longer there?

christianbundy:
@mixmix Any chance it's in `git reflog`? I've lost *lots* of work to Git but almost always found it in the ref log (although I thought it was "re-flog" for ages...).

mixmix:
Disabled indexing for the moment.
I think I can get it down to one click of next on launch... Also I might be
able to get it out auto skip this stuff but want to build later.
ie AHOY=true isn't there any more.

As I said mainly intetested in whether ripping out config + electron is
upsetting for anyone.



Christian I didn't commit it and ran git checkout .

arj03:
Testing the new create user functionality, seems to work. It created a new folder and I copied my config over just to be sure everything should be the same, but starting up I get this error:

```
Error: shs.server: client hello invalid (phase 3). they dailed a wrong number - they didn't have our public key
    at abort (/home/chrx/dev/patchbay/node_modules/ssb-server/node_modules/secret-handshake/protocol.js:82:45)
    at Object.cb (/home/chrx/dev/patchbay/node_modules/ssb-server/node_modules/secret-handshake/protocol.js:95:20)
    at drain (/home/chrx/dev/patchbay/node_modules/ssb-server/node_modules/pull-reader/index.js:39:14)
    at more (/home/chrx/dev/patchbay/node_modules/ssb-server/node_modules/pull-reader/index.js:55:13)
    at /home/chrx/dev/patchbay/node_modules/ssb-server/node_modules/pull-reader/index.js:66:9
    at /home/chrx/dev/patchbay/node_modules/ssb-server/node_modules/pull-reader/index.js:20:7
    at drain (/home/chrx/dev/patchbay/node_modules/ssb-server/node_modules/stream-to-pull-stream/index.js:141:18)
    at Socket.<anonymous> (/home/chrx/dev/patchbay/node_modules/ssb-server/node_modules/stream-to-pull-stream/index.js:150:5)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
```

I think the settings should be tied to the specific config/identity, so for one I might choose only connect to my pub and for another I might have connect to all. Also with this I could do separate css for these different identities so that one of them was say pink background to make it easier to distinguish. So no, not upsetting, this the way to go I think.

mixmix:
hey @arj03 wasn't sure what you meant by this : 
> I think the settings should be tied to the specific config/identity

when you pick an identity 
- e.g. `ssb` it loads `~/.ssb/config` and lets you edit (it persists it for you)
- e.g. `naarm` it loads `~/.naarm/config` and lets you edit that

Which "settings" are you talking about? With the tor and pub settings, as far as I can tell they are really "config" and I'd ultimately like to add a feature where you can e.g. click a "tor only" button and it auto-edits the config for you. Same with pubs. I've not implemented this yet though ... it really needs doing before this is all merged otherwise it's a feature regression for people. SO MANY things. wow

arj03:
@mixmix I meant the settings in patchbay. Yeah, we might do the config change, I could have a look at that if we can agree on a strategy. But I was also referring to the css styling you can do, not sure if that fits into the normal config? Maybe a specific patchbay config file?

Also I just wanted to put the error out there, my hunch is that its because sbot is using a different id that you just added but havn't looked at the code yet.

Also this looks great! :)

arj03:
I pushed up two minor spelling fixes, hope you don't mind ;)

I'm one of those people with a custom edited config where I have commented out some things with /**/, so doing straight up JSON.parse on that file crashes. RC (as ssb-config does) can handle those files, but if we write what is read we'll end up truncating these comments. 

For now I changed my config to test the rest of the stuff. I see a nice loading screen, excellent! My tor and pub hub settings are reset, not a big problem. But the config is not saved on restart. Fixed that ;) Also fixed saving tor settings.

arj03:
Was trying to debug the new config error, npm run dev is broken with this new setup? Doesn't seem to help to point it to exports.js instead of main.js in package.json.

arj03:
That error from creating a new user is really wierd. I tried copying my conn.json file over and I can see in /network that it is doing connections to other hosts. Also the error is from net, and we should be using socket to connect to the local sbot dæmon. Sorry I don't have more time to debug for now, its already waaay to late here :-)

mixmix:
Thanks for those fixes! Especially nice catches with cloneDeep and parseInt.

Well one thing we could do is use json5 to parse, there's a good chance it
handles comments.
I don't really know how one preserves comments.

Can I ask what the deeper need is - Why do you need comments?

What's the remaining error you're experiencing? That one where a client is
still trying to connect with wrong public key mm is there a computer on the
same network that thought it could contact @arj at a local ip, but is
instead erroring when it is getting @arj2?

mixmix:
> Was trying to debug the new config error, npm run dev is broken with this new setup? Doesn't seem to help to point it to exports.js instead of main.js in package.json.

@arj03  `npm run dev` will perhaps work if it points just at `ui.js`, but you'll need to add a line in that file so that it knows if it's not being required to go ahead an execute `Start(config)`, where that config is probably just the default `ssb-config` generated config.

I'm not going to focus on that for the moment. Also note that if you do this you won't get any of the ahoy launcher stuff - that's because ahoy manages several UI's - the ahoy launcher, and is passed Patchbay's UI.

arj03:
npm run dev is fine for now at least for now, just wanted to put it out there. I've actually used npm start for the past month anyway 😮 

The need for comments was mostly for when I was testing different things directly on sbot, then I would comment things out that way so I didn't need multiple files. So probably not such a big deal except if someone does that patchbay fails to launch, at least we won't overwrite their config. I'm not sure what the best solution is, patchbay is I think the first ssb software that actively modifies the config.

I know I'm the odd duck here with the small things listed above :)

As for the last error I'll try see if I can figure it out today. It is just starting up patchbay with a different profile than my main. No other machines are involved.

arj03:
The error seems to come from my net incoming connection. The strange thing is that socket & ws is working fine. If I just have a config with only those two, I don't get any errors. I wonder why net seems to have a different key than the other two connections.

mixmix:
@arj03 I fixed that bug with the config comments - the only gotcha is if you use the ssb-ahoy editor you will loose your comments...

mixmix:
HERE WE GO

PR #317, opened 2019-02-24
fix device scope

dominictarr:
`local/private` scope is the local network, i.e. wifi. _not_ localhost. that is the `device` scope.

A unix socket can only be connected to within the device.

This will fix https://github.com/ssbc/ssb-server/issues/640

I'm not really happy with the scope names. It seems like a confusion as to meaning of the names lead to this bug..

I'd kinda lead towards "device" "nearby" "global" or something. get away from networking terms?

### DIFF
--- a/config.js
+++ b/config.js
@@ -33,7 +33,7 @@ function addSockets (config) {
     config,
     {
       connections: {
-        incoming: { unix: [{ scope: 'local', transform: 'noauth', server: true }] }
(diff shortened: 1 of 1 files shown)


dominictarr:
also, appologies for using github edit to make the PR. I had just enough time to do that.

christianbundy:
Oops, looks like I forgot to PR cd91f627f4e517409efe0ca1fd453bd9947b65f9.

Merging!

PR #318, opened 2019-03-09
replace gossip plugin with new rewrite

staltz:
I've been working on the gossip plugin refactor/rewrite. `ssb-legacy-conn` is essentially the same functionality as `ssb-gossip`, but with its internals reworked with a new stack of libraries (notably, `ssb-conn-db` migrates the file `gossip.json` to `conn.json`). I'm using `ssb-legacy-conn` also in Manyverse.

Try out this PR branch before merging. Everything should functionally be the same, the one difference is that `gossip.json` is *read* but never *updated*, only `conn.json` is updated from now onwards. I'll be updating `ssb-legacy-conn` gradually until its internals are very easy to work with. Then it'll be easy to create alternative plugins, e.g. `ssb-better-conn`.


mixmix:
hey just saw this. Thanks for posting this @staltz 

I'm in the middle of a cross-platform nightmare. I look forward to trying this and hopefully getting it in production. If anyone else tests this before me and finds it good feel free to merge.

Note if what I'm working on currently gets merged we'll need to add this to `ssb-ahoy` too!

staltz:
Just updated this PR so it uses an even newer ssb-legacy-conn. Again, this should be without any breaking changes, and Manyverse is also using this with no problems. Should be an easy PR to test, but I'm not putting any pressure, take your time. :)

arj03:
Testing this. Found a missing dependency bug in ssb-invite. There is also something funky in /network. I can see that I'm connected to 3 pubs, but none show up under remote peers.

arj03:
@staltz looks like patchbay gets remote peers [here](https://github.com/ssbc/patchbay/blob/9b0b78d08222f41f2d21ddae8b55449a25a0784c/app/page/network.js#L166). This comes from [patchcore](https://github.com/ssbc/patchcore/blob/63f89f6d4f6475a287ab16681374966538fdf0da/sbot.js#L105). Seems like state is never set on the peers?

staltz:
@arj03 Very good find, thanks! I released ssb-legacy-conn@1.0.17 and updated this PR. It should have `.state` on the peers now.

arj03:
Can confirm that this now works. Have been running this since yesterday, so it connected and downloaded todays bulk. Also happy that this plays nice with ssb-friend-pub.

@staltz I was skimming the source just out of curiosity and can see references to DHTs. What is needed to support that in patchbay? Also the code is quite readable which is really nice, especially considering I havn't done any typescript coding.

staltz:
Whoo! Thanks for merging!

ssb-legacy-conn supports DHT connections, that's one of the few deviations it has from ssb-gossip. You can take a look at Patchwork's PR to add DHT invites: https://github.com/ssbc/patchwork/pull/862/files it's mostly correct, but use ssb-legacy-conn instead of `@staltz/sbot-gossip`. Also ping me in the PR you make for Patchbay so I can help review.

mixmix:
Nic ework all, Thanks both!

PR #321, opened 2019-03-25
update ssb-legacy-conn to 1.0.23

staltz:
This fixes some bugs (specifically for LAN sync) that brings behavior of this library closer to ssb-gossip

cc @arj03


arj03:
Tested this tonight and works fine. Merging. Thanks!

PR #325, opened 2019-04-19
Add book notifications

arj03:
I wanted to keep the code for pulling the notifications seperate, so it could be tested. So this is mostly a pull-merge :)

scuttle-book-notifications


arj03:
If I had just added this in the first version. The new structure is so much nicer to work with though. Tests :heart:

PR #326, opened 2019-04-23
Support sharing files using dat

arj03:
This should be ready to go I think. What it does is use the [dat-shared-files](https://github.com/arj03/dat-shared-files) module I created to make it easier to share files with dat. The integration into patchbay opens dat links as external links, so should open in beaker browser.

Would love for someone to give this a spin before merging :-)

### DIFF
--- a/app/sync/initialise/dat.js
+++ b/app/sync/initialise/dat.js
@@ -3,11 +3,6 @@ const datSharedFiles = require('dat-shared-files/lib')
 
 exports.gives = nest('app.sync.initialise')
 
-exports.needs = nest({
-  'router.sync.router': 'first',
-  'app.html.tabs': 'first'
-})
-
 exports.create = function (api) {
   return nest('app.sync.initialise', datShare)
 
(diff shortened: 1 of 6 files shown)


mixmix:
Reviewing this now then smoke testing it!

app/sync/initialise/dat.js
```diff
@@ -0,0 +1,19 @@
+const nest = require('depnest')
+const datSharedFiles = require('dat-shared-files/lib')
+
+exports.gives = nest('app.sync.initialise')
+
+exports.needs = nest({
+  'router.sync.router': 'first',
+  'app.html.tabs': 'first'
+})
```
mixmix: :wastebasket:  these `needs` are not used

mixmix: fixed

mixmix:
This seems to be working great.

I've added a little styling (and fixed some embarassing old code of mine!). It now looks like this : 

image

PR #332, opened 2019-05-13
update for ssb-server@15 use ssb-* plugins directly, ssb-server/plugins/* no longer supported

dominictarr:
we are about to remove ssb-server/plugins directory, so these lines need to change.
the goal here is to make ssb-server into _just a distribution_. It doesn't have any complex code anymore.


mixmix:
thanks @dominictarr - I'm currently tying to get a stable major release out. As usual, that's taken a lot more time than expected - including problems with `ssb-ws` and `ssb-invite`, which I've had to lock at old versions for the time being.

Keen to have a look at this in the next pass. In particular, getting to remove the shrinkwrap could be super sweet because there is massive amounts of duplication of `sodium-native` binaries that I have been totally unable to prune.

server.js
```diff
@@ -6,12 +6,12 @@ var electron = require('electron')
...
-  .use(require('ssb-server/plugins/unix-socket'))
-  .use(require('ssb-server/plugins/no-auth'))
-  .use(require('ssb-server/plugins/onion'))
-  .use(require('ssb-server/plugins/local'))
+  .use(require('ssb-master'))
+  .use(require('ssb-logging'))
+  .use(require('ssb-unix-socket'))
+  .use(require('ssb-no-auth'))
+  .use(require('ssb-onion'))
+  .use(require('ssb-local'))
```
christianbundy: Oh, I was thinking that ssb-server *as a distro* should be including these by default, maybe I didn't understand correctly?

mixmix:
I think we'll really start to feel the shape of this when we separate the
opinionated CLI from the requirable server

dominictarr:
I think the problem here is that ssb-server is intended to be a _distribution_, but really, patchbay has it's own distribution. you should skip depending on ssb-server, and depend on secret-stack and build up all the plugins that you choose, as distribution maintainer.

Also you should run the tests for that combination of plugins, before releases. TODO: https://github.com/ssbc/ssb-server/issues/650

christianbundy:
This is sort of weird but it works.

ssb-ahoy depends on ssb-server *and* the individual plugins used, so `package.json` includes both. I've run `npm dedupe` so these point to the same files, but it's still a slightly unconventional config to ensure that it doesn't break ssb-ahoy.

The biggest change here from my perspective is that it means Patchbay will start supporting deleted messages in the database. In Patchwork `master` we're currently deleting all messages from blocked feeds, which means that when I switch to Patchbay it crashes because it uses an old ssb-db that doesn't support deleted messages. To be super explicit, this PR won't make Patchbay start deleting anything, it just makes it so that it understands when *other* clients delete messages from the shared database.

arj03:
Seems as if this is in master now.

christianbundy:
Cool!

PR #338, opened 2019-05-23
WIP - slider for hops!

mixmix:
@marylychee I took the liberty of adding you to the SSBC org (you may have to accept), and pulling your branch into the repo here so that I could start a WIP PR that we could look at and more easily discuss and jam on

### DIFF
--- a/app/html/settings/friend-hops.js
+++ b/app/html/settings/friend-hops.js
@@ -26,15 +26,14 @@ exports.create = function (api) {
     hops.initial = hops()
 
     watch(hops, hops => {
-      const intHops = parseInt(hops)
-      updateConfig(intHops)
+      updateConfig(hops)
     })
 
     return {
       title: 'Friend Hops',
       body: h('FriendHops', [
         h('div.description', [
-          'What you replicate (store a local copy of) is based on how many "hops" you replicate. If you replicate out to 1 hop, you are replicating the people you follow, at 2 hops, it is your follows and people they follow.'
+          'What you replicate (store a local copy of) is based on how many "hops" you replicate. If you replicate out to 1 hop, you are replicating the people you follow, at 2 hops, it is your follows and people they follow. Play with the slider to see this visualised in the graphic below!'
         ]),
         h('div.slider', [
           h('datalist', { id: 'friends-hop-datalist' }, [
@@ -49,11 +48,15 @@ exports.create = function (api) {
             min: 0,
             max: 3,
             value: hops,
-            'ev-change': (ev) => hops.set(ev.target.value)
+            'ev-change': (ev) => hops.set(parseInt(ev.target.value))
           })
         ]),
         h('div.alert',
-          { style: { opacity: computed(hops, (_hops) => (_hops === hops.initial) ? 0 : 1) } },
+          {
+            style: {
+              opacity: computed(hops, (_hops) => (_hops === hops.initial) ? 0 : 1)
+            }
+          },
           [
             h('i.fa.fa-warning'),
             ' please restart patchbay for this to take effect'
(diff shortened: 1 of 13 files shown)


app/html/settings/friend-hops.js
```diff
@@ -0,0 +1,84 @@
...
+    hops.initial = hops()
+
+    watch(hops, hops => {
+      const intHops = parseInt(hops)
+      updateConfig(intHops)
+    })
+
+    return {
+      title: 'Friend Hops',
+      body: h('Friends', [
```
mixmix: make this `h('FriendsHops', ...)`

This is shorthand for `h('div.FriendsHops', ...)  which makes a div with class `.FriendsHops`. If you name it this it will match the MCSS class you made and things should line up !

app/html/settings/friend-hops.js
```diff
@@ -0,0 +1,84 @@
...
+          if (_hops === hops.initial) return
+          return h('div.alert', [
+            h('i.fa.fa-warning'),
+            ' please restart patchbay for this to take effect'
+          ])
+        }),
+      ])
+    }
+
+    function updateHops (hops) {
```
mixmix: this isn't used / isn't relevant here. ... we could add a thing which calculates how many people you're goingto be replicating at different hops ... but that feels like "later"

app/html/settings/friend-hops.js
```diff
@@ -0,0 +1,84 @@
...
+    watch(hops, hops => {
+      const intHops = parseInt(hops)
+      updateConfig(intHops)
+    })
+
+    return {
+      title: 'Friend Hops',
+      body: h('Friends', [
+        h('div.description', [
+          'Specify whose content you replicate, and thus the size and shape of your network.'
```
mixmix: Probably we need some helper test explaining what 0, 1, 2, .. means (1 = just people you followed, 2 = people you follow and their friends). That or a link out to [6 degress of kevin bacon](https://en.wikipedia.org/wiki/Six_Degrees_of_Kevin_Bacon)

mixmix:
SUMMARY: looks great. Couple of changes needed that I've commented on. Happy to discuss any parts of this which were too dark-magic (yes there are some, and yes we regret some of those)

app/html/settings/friend-hops.js
marylychee: oh right, thank you!

app/html/settings/friend-hops.js
marylychee: I see, I wasn't sure whether to keep it or not. Ok, sounds good

marylychee:
thanks @mixmix this was fairly fun! easy enough changes, and first time having a clear WIP PR to jam on together. Hope it's all good to go now :) 

however in development, I think I messed up something with my ssb so that I can't use ssb right now on any of the clients. 😅 which makes it harder for me to try troubleshoot/reach you guys to ping for answers. Maybe the indexing got screwed up?

On patchwork I just get infinite load, and on patchbay - the ahoy launch page is fine, but when I click to launch the app, in my terminal I'm getting a load of these bad boys: 
```
Error {}
Error {}
Error {}
Error {}
Error {}
RangeError: Maximum call stack size exceeded
```

and then 
```
when trying to retrive:typ,post,1558607339615at since:779741888
```

🤷🏻‍♀️  is there an ssb troubleshoot instructions for when you can't access ssb?

mixmix:
hey @marylychee - good point, there should be a guide for "when your setup is messed up" on e.g. `scuttlebutt.nz` ... I've helped a few people unmess up their setups and have had a bunch of help when I've done something stupid. Maybe this could go into the village-tracker ? 

---

spun this up. I'm gonna tweak the styles a bit because they're a bit cramped at the moment : 
image

code is looking good

Powersource:
Does anyone have a screenshot of that dynamic svg in the app? Would be a cool addition to https://decentpatterns.xyz/library/social-radius-slider/

Issue #354, opened 2019-07-18
Migrate from Gossip to CONN


staltz:
Here's a big issue that covers a lot of TODOs in order to move away from ssb-gossip and adopt ssb-conn. Feel free to edit this list if you come across some other task related to gossip.

- [ ] **Patchcore REQUIRED**
  - `patchcore/sbot.js` should replace `sbot.gossip.peers` (a `sync` api) with `ssb.conn.peers` (a `source` api), should pull.drain that and update `localPeers` and `connectedPeers`
- [ ] **Patchcore REQUIRED** 
  - `patchcore/sbot.js` should expose a new api called `stagedPeers`, similar to `localPeers` and `connectedPeers`, based on `ssb.conn.stagedPeers` (a `source` api)
- [ ] **Patchcore** OPTIONAL 
  - `patchcore/sbot.js` should replace `gossipConnect` with `connConnect`, replace `sbot.gossip.connect` with `sbot.conn.connect`
- [ ] **Patchbay REQUIRED** 
  - patchbay's Network page should also observe `stagedPeers`, display them, and allow the user to "approve" a staged peer, which should call `patchcore.connConnect` and then publish a new follow message
- [ ] **Patchbay REQUIRED** 
  - patchbay should use ssb-lan instead of ssb-local
- [ ] **ssb-ahoy** OPTIONAL 
  - `ssb-ahoy/views/replication` should replace `server.gossip.connect` with `server.conn.connect`
- [ ] **ssb-invite** OPTIONAL 
  - `ssb-invite` should replace `server.gossip.add` with `server.gossip.remember`
- [ ] **ssb-replicate** OPTIONAL 
  - `ssb-replicate` should replace `ssbServer.gossip.peers` (a `sync` api) with `ssbServer.conn.peers` (a `source` api), pull.filter that to have only `type==='lan'` and truthy `.key`, pull.drain that, and `request()` replication
- [ ] **ssb-ebt** OPTIONAL 
  - `ssb-ebt/index.js` should remove the `sbot.gossip.disconnect` call because ssb-conn's scheduler automatically makes sure that blocked peers will never be connected to
- [ ] **ssb-device-address** OPTIONAL 
  - `ssb-device-address/index.js` should replace `sbot.gossip.add/remove` with `sbot.conn.remember/forget`, notice the TODO comment in ssb-device-address, ssb-conn supports multiserver addresses
- [ ] **ssb-friend-pub** OPTIONAL 
[...]

PR #355, opened 2019-07-29
ssb-conn

mixmix:
this branch is about getting staltz's `ssb-conn` into play!
So far seems to be happily running

**Things still needed**
- some time testing this to make sure it's fast / stable / behaving with memory
- make all the must-have changes listed here https://github.com/ssbc/patchbay/issues/354 !
- figure out how to prevent the peerDependency of `ssb-conn` (`secret-stack`) from exploding the install / rebuild


staltz:
Stap it bot

arj03:
@mixmix what is the status here? It seems as if latest master if running ssb-conn?

staltz:
@arj03 I guess this issue tells everything? https://github.com/ssbc/patchbay/issues/354

As far as I see, patchcore needs to be updated and then included in patchbay, among other fixes. The list in that issue is so thorough, that it's just a matter of implementing exactly each item. Some of those changes are actually basically the same changes that I did in Patchwork, because Patchwork had a copy-paste of some patchcore code instead of an actual dependency.

PR #356, opened 2019-08-11
Use ssb-lan instead of ssb-local for better privacy

arj03:
This uses ssb-lan. 

I noticed that ssb-ahoy also includes ssb-local in ./lib/build-plugins.js, so that probably also needs updating?

Tested IRL with @Powersource and @cblgh

### DIFF
--- a/app/sync/initialise/lan-discovery.js
+++ b/app/sync/initialise/lan-discovery.js
@@ -1,0 +1,32 @@
+const { onceTrue } = require('mutant')
+const nest = require('depnest')
+const pull = require('pull-stream')
+
+exports.needs = nest({
+  'sbot.obs.connection': 'first',
+  'config.sync.get': 'first'
+})
+
+exports.gives = nest('app.sync.initialise')
+
+exports.create = function (api) {
+  return nest('app.sync.initialise', lanDiscovery)
+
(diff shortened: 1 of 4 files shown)


staltz:
Nice that you're also going to use ssb-lan! If this is for "better privacy", note that currently ssb-lan emits **both** the legacy style (hence compatible with other peers still using ssb-local) as well as the new style. Using the legacy style means we have no privacy, so basically ssb-lan will only provide privacy once we stop using the legacy mode. If you want, we could add an option in ssb-lan to disable legacy broadcast, but this comes with the cost that we won't discover other peers using ssb-local (e.g. Patchwork users).

Also, could this PR be done at the same time as ssb-conn is used in Patchbay? This way you wouldn't have to add any new code in `app/sync/initialise/`, just `.use(require('ssb-lan'))` and ssb-conn will handle the discoveries for you.

arj03:
Hah, I just came to the same conclusion about ssb-conn as you can see in the other issue :)

And yeah, I'm thinking of rolling this in with legacy still enabled and once patchwork is updated we can disable it I think.

staltz:
Stop bot

arj03:
This is no longer relevant as latest master includes ssb-lan

PR #359, opened 2019-10-05
Comments please:

lancew:
This set of commits provides a setting to save open tabs, o that when you close and reopen patchbay it retains the tabs you had open when you closed it.

Improvements, suggestions, mentoring on better approaches welcomed.

### DIFF
--- a/app/sync/go-to.js
+++ b/app/sync/go-to.js
@@ -8,7 +8,9 @@ exports.needs = nest({
   'history.obs.store': 'first',
   'history.sync.push': 'first',
   'router.async.normalise': 'first',
-  'router.async.router': 'first'
+  'router.async.router': 'first',
+  'settings.obs.get': 'first',
+  'settings.sync.set': 'first'
 })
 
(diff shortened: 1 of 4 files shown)


app/sync/go-to.js
```diff
@@ -59,6 +61,24 @@ exports.create = function (api) {
         page.id = page.id || locationId
         tabs.add(page, !openBackground, split)
 
+        // Save Tabs: If enabled then add it to default tabs.
+        const saveTabs = api.settings.obs.get('patchbay.saveTabs', false)
+        if (saveTabs()) {
+          const openTabs = api.settings.obs.get('patchbay.openTabs', [])
+          var _tabs = openTabs()
```
jedahan: If using const, why not use `let` instead of `var` to be a tiny bit more strict regarding variable access?

lancew: Thx... will do.
I "think" I can use api.settings.sync.get rather than obs.get too

PR #362, opened 2019-12-18
Remove ssb local

arj03:
Just minor bumps


christianbundy:
Yep, looks right to me. Thanks for the patch!

PR #368, opened 2020-03-02
README: drop .git from clone url

edrex:
this avoids .git suffix in default working copy path.


arj03:
Looks good. Thanks @edrex :)

